PR #3, opened 2024-10-31
Html css/soheib

siderdk:
first homework submission + finishing the first lesson.


html-css/week1/homework01/index.html
```diff
@@ -0,0 +1,70 @@
...
+                <div class="navcard"><a href="#about">About</a></div>
+                <div class="navcard"><a href="#skills">Skills</a></div>
+                <div class="navcard"><a href="#portfolio">Portfolio</a></div>
+                <div class="navcard"><a href="#contact">Contact</a></div>
+            </div>
+        </header>
+        <section class="contwrapper">
+           <div class="contcard" id="about">
+            <p>I'm <b id="name">Soheib</b></p>
+                <p><img id="profilepic" src="./assets/577A1074 4.jpg" alt="profile picture">
```
Khandohii, line 29: Bad value ./assets/577A1074 4.jpg for attribute src on element `` Illegal character in path segment: space is not allowed.

html-css/week1/homework01/styles.css
Khandohii, line 1: The code is clear and understandable, but I would remove unnecessary comments, and also remove unnecessary line breaks.

Also, for scalability, it is better to use class selectors instead of IDs, since only one unique ID is allowed per page.

Khandohii (approved):
Well done @siderdk 👍 

Left some comments to fix, and then can merge it.

html-css/week1/homework01/index.html
siderdk: Thank you very much for the advice. I have implemented the change

html-css/week1/homework01/styles.css
siderdk: Many thanks for the valuable feedback. 
I removed unnecessary linebreaks and comments. 
I also remove IDs except those connecting sections to the navbar, to ensure smooth scroll. Is that ok? 
I found that there is a way to do in JS but I'm not there yet.

Khandohii: Good job @siderdk,
Yes, that's OK to use IDs for the anchor of a navbar. You'll learn it later, how to interact with JS and smooth navigation.

Khandohii: Now you can merge this branch

siderdk: Thank you so much

PR #12, opened 2024-11-01
html-css homework

pahanan:


Khandohii (changes requested):
It looks good 
I would add this meta tag into the HEAD of the HTML for better responsibility and visibility on different mobile devices:
`<meta name="viewport" content="width=device-width, initial-scale=1, maximum-scale=1">`

html-css/week1/script.js
Khandohii, line 1: This file is not used in the project, so it can be deleted.

Khandohii (changes requested):
Well done 👍 
There are a couple of corrections left

Khandohii (approved):
Well done, Andrii! 👏

PR #15, opened 2024-11-03
Homework_First_Week

jalilhu:
initial commit
Replicating a profile website


html-css/week1/index.html
```diff
@@ -0,0 +1,185 @@
...
+    <link rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/font-awesome/6.0.0-beta3/css/all.min.css">
+    <title>S J H </title>
+</head>
+<body>
+    <header>
+        <nav>
+            <img id="profile-pic" src="./images/me.jpeg" alt="profile-sjh"/>
+            <div class="nav-container">
+                
+                <div id="nav-home" class="nav-item">
```
NVedsted, line 18: Excellent re-use of CSS through `nav-home`! 🚀 However, we should strive for IDs to unique identify elements thus having multiple elements with the same ID is not good. There is also issues with this for `square`, `pro-item`, and `test-item`.

Can you think of an alternative method to reuse styling that avoids elements sharing IDs? 😄

html-css/week1/index.html
```diff
@@ -0,0 +1,185 @@
...
+                </div>
+            </div>
+        </section>
+
+   </main>
+   <footer>
+    © Jalil Hu      <a href="#">Privacy Policy</a> | <a href="#">Legal Notice</a>
+   </footer>
+    <script>
+        function scrollToAboutme() {
```
NVedsted, line 200: Already some JavaScript making its way into the code! 😮 In this case, we could accomplish the same using pure CSS, can you figure out how this would work? (hint: look into `scroll-behavior`)

html-css/week1/styles.css
```diff
@@ -0,0 +1,1240 @@
+@import url('https://fonts.googleapis.com/css2?family=Roboto:ital,wght@0,100;0,300;0,400;0,500;0,700;0,900;1,100;1,300;1,400;1,500;1,700;1,900&display=swap');
```
NVedsted, line 1: This files contains a few, albeit minor, issues, and I would recommend you running it through a CSS validator (perhaps [this one](https://jigsaw.w3.org/css-validator/) which finds 15 errors and 6 warnings). 🔍

NVedsted (changes requested):
Hey!

Good work with this page! 🎆 You certainly got around to do a lot of things and it turned out nice. 👍 

I have left a few comments regarding some issues to be addressed. Generally, the formatting of files is a bit inconsistent. I'd suggest reformatting your files either with your editor or a tool. ([VS Code can format files for you!](https://code.visualstudio.com/docs/editor/codebasics#_formatting)) 📄 

Good luck with it and let me know if you have any questions! 😄

html-css/week1/index.html
jalilhu: I should have used class. 
thanks for pointing it out.

html-css/week1/styles.css
jalilhu: Great tool.
find and removed all the issues. 
thanks  a lot.

html-css/week1/index.html
jalilhu: <a  tag and with href property that navigates to given id

PR #20, opened 2024-11-04
Git-week1/farzaneh

FarzanehAhmadi:


margaritluch (commented):
Hi Farzaneh
Everything looks fine. The only thing that need to be change is the branch name . It should be git-week1/farzaneh according to the requirements.

PR #21, opened 2024-11-04
add three files as per homework instruction nr.6-7-8

siderdk:
This is the first exercice. 
If you're a vegetarian, make copy the recipes ;)

### DIFF
--- a/git/week1/countries.txt
+++ b/git/week1/countries.txt
@@ -1,0 +1,3 @@
+Greenland
+Japan 
+Palestine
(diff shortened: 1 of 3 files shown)


RezaShahrestani (changes requested):
Dear Soheib,
File changes and names looks good, also your commit message looks okay.
But as instructions said you need to commit each file in a separate commit.

Now for practicing this, Edit each file once and commit it with proper message.

siderdk:
Thank you Reza for the feedback. 
I will pay more attention to the instructions next time. 
Now to the fixes :)

PR #23, opened 2024-11-04
three files were added

jalilhu:


Hadis-jamali (commented):
You had to have 3 commits for each file but you have just one commit, try to change the files and commit each file again.
read homework assignments and do them one by one

jalilhu:
Thanks Hadis for reviewing
I just deleted all 3 files and followed the assignment correctly. thanks

PR #26, opened 2024-11-05
git_week1_homework_1

jianxinz233:


margaritluch (commented):
Hi Jianxin
Can see you have all the necessary files committed. Well done. The only thing that needs to be changed is the branch name. According to the requirements it should be git-week1/jianxin

PR #27, opened 2024-11-05
Git week1/aminata

Aminata00:
git-week1/Aminata pull request exercise 1


margaritluch (approved):
Hi Aminita
I can see you have committed all the necessary files. Well done. The only thing you need to change is the branch name. It should be git-week1/aminita according to requirements.

PR #36, opened 2024-11-07
html-css/juani

PAINj9:
HTML Homework


html-css/portfolio/script.js
```diff
@@ -0,0 +1,192 @@
...
+                nextText.classList.add('current');
+            }, 1000); // Tiempo de transición de texto
+
+            currentIndex = (currentIndex + 1) % textArray.length;
+        }
+
+        setInterval(changeText, 2500); // Intervalo de cambio de texto
+    }
+
+    let currentSlide = 0;
```
AngelHenriettaAboah: I see you are using javascript to build your portfolio so I will leave comments based on this. Best practice was using Html and css to get acquainted with using them within a project when called for. 
To prevent global scope pollution, it would be nice to encapsulate variables within their respective functions. An example could be encapsulating "currentSlide"  and  "skillSlideIndex" within their own functions.

html-css/portfolio/script.js
```diff
@@ -0,0 +1,192 @@
...
+function prevSkillSlide() {
+    skillSlideIndex--;
+    showSkillSlides();
+}
+
+showSkillSlides();
+
+
+// Cambiar entre modo claro y oscuro al usar el interruptor
+const themeToggle = document.getElementById('theme-toggle');
```
AngelHenriettaAboah, line 200: Add checks to ensure elements like textWrapper, themeToggle, and rotatingText exist before proceeding with animations or effects to prevent JavaScript errors.

AngelHenriettaAboah (changes requested):
Great Effort! Left comments already.

PAINj9:
I did those changes, thanks :)

PR #42, opened 2024-11-07
git-week1

PAINj9:
git-week1


margaritluch (commented):
Hi Juan
You only needed to change 3 files. There is no need to change 18. You also need to change the branch name to git-week1/juan according to requirements.

PR #59, opened 2024-11-11
 javascript1 week1 homework(AndriiP)

andrpavl:
Finished homework to JS-1st module, 1st week


javascript/javascript1/week1/exercise1.js
```diff
@@ -0,0 +1,8 @@
+// Exercice 1 Age-ify (A future age calculator)
+
+const yearOfBirth = 1988;
+let yearFuture = 2047;
```
LucasIversen: we can make this a `const` since we're not changing the value

javascript/javascript1/week1/exercise2.js
```diff
@@ -0,0 +1,14 @@
+// Exercise 2 Goodboy-Oldboy (A dog age calculator)
+
+const dogYearOfBirth = 2005;
+let dogYearFuture = 2024;
+const dogYear = (dogYearFuture - dogYearOfBirth) * 7;
+let shouldShowResultInDogYears = true;
+
+shouldShowResultInDogYears
```
LucasIversen, line 8: Nice use of the ternary operator! I'd give prettier a go for formatting ;)

javascript/javascript1/week1/exercise3.js
```diff
@@ -0,0 +1,45 @@
+//   Exercise 3 Housey pricey (A house price estimator)
+
+const calcVolumeInMeters = (width, depth, height) => {
```
LucasIversen, line 3: Niice!

javascript/javascript1/week1/exercise4.js
```diff
@@ -0,0 +1,37 @@
...
+	"Deal",
+	"Future",
+	"Life",
+	"Day",
+	"Adventure",
+	"Fairytale",
+	"Prank",
+];
+
+const startupName = firstWords[randomNumber] + " " + secondWords[randomNumber];
```
LucasIversen: The exercise is to match the two words at ramdom. By only having one random index accessor we always get the words in parallel. Make two random numbers, and use only one per array to match at random

LucasIversen (changes requested):
Very good solution Andrii. Just need a few adjustments

LucasIversen (approved):
No need to make more changes, but please note that I'd like `dogYearFuture`and `shouldShowResultInDogYears` in exercise 2 to be const as well..

PR #66, opened 2024-11-13
Javascript javascript week1 aminata

Aminata00:
javascript homework

### DIFF
--- a/javascript/javascript1/week1/ageify.js
+++ b/javascript/javascript1/week1/ageify.js
@@ -1,0 +1,4 @@
+const yearOfBirth = 1987;
+const yearFuture = 2027 - yearOfBirth; 
+const age = "You will be " + yearFuture + " in 2027.";
+console.log( age )
--- a/javascript/javascript1/week1/ageify.js
+++ b/javascript/javascript1/week1/ageify.js
@@ -1,4 +1,5 @@
 const yearOfBirth = 1987;
-const yearFuture = 2027 - yearOfBirth; 
-const age = "You will be " + yearFuture + " in 2027.";
+const yearFuture = 2027
+const year = yearFuture - yearOfBirth; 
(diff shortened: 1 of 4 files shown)


javascript/javascript1/week1/ageify.js
```diff
@@ -0,0 +1,4 @@
+const yearOfBirth = 1987;
+const yearFuture = 2027 - yearOfBirth;
```
LucasIversen: We'd like to have the future year in a variable of it's own, so

`const futureYear = 2027`

javascript/javascript1/week1/eznamey.js
```diff
@@ -0,0 +1,5 @@
+const firstWords = ["fabulous", "amazing", "awesome", "great", "wonderful", "genius", "creative", "incredible", "stunning"];
+const secondWords = ["lab", "studio", "hub", "works", "forge", "factory", "nest", "hive", "vault", "collective"];
+const randomNumber = Math.floor(Math.random() * 10);
```
LucasIversen: Use two random numbers to access the arrays. With this solution the words will always come out in parallel, so if the first word is Awesome, the second one will always be hub.

javascript/javascript1/week1/houseypricey.js
```diff
@@ -0,0 +1,17 @@
+    
+const house1 = 8 * 10 * 10 * 2.5 * 1000 + 100 * 300;
```
LucasIversen: Can we put the width, height, depth values and garden size values in their own variables like:

`const height = 500;` ect.

Makes the code more readable :)

LucasIversen (changes requested):
Good solution Aminata. Just need a few tweaks, and we're good

LucasIversen (approved):

PR #69, opened 2024-11-14
Javascript javascript1 week1/soheib

siderdk:
I tried to git each exercise a frontend part, but I only succeeded with the first one, and then we are all sick at home. so now, I'm just hoping I'll be fine to attend on Sunday! 
Thank you in advance for your help dear Mentor :)


javascript/javascript1/week1/homework_soheib/ezNamey.js
```diff
@@ -0,0 +1,6 @@
+let firstWords = ["One", "New", "Di", "All", "The", "Dan", "A", "Obsidian", "Cloud", "365"];
+let secondWords = ["Tech", "Jysk", "Dev", "Sol", "Suite", "In", "AI", "Click", "Soft", "Light"]
+let randomIndex1 = Math.floor(Math.random() * firstWords.length);
```
LucasIversen: these can all be const since we're not changing them after initialization

javascript/javascript1/week1/homework_soheib/goodboy_oldboy.js
```diff
@@ -0,0 +1,12 @@
+let dogYearOfBirth = 2014;
+let dogYearFuture = 2025;
+let dogYear = 7;
+let shouldShowResultInDogYears = false;
+
+if (shouldShowResultInDogYears === true) {
+    let result = (dogYearFuture - dogYearOfBirth) * dogYear;
+    console.log(`Your  dog will be ${result} human years old in ${dogYearFuture}`)
+} else if(shouldShowResultInDogYears === false){
+    result = dogYearFuture - dogYearOfBirth;
```
LucasIversen: have you tested this? This should throw an error as result has not been initialized as a variable. Remember that the code in line 7 and 8 is not run if shouldShowResultInDogYears is not true :)

javascript/javascript1/week1/homework_soheib/houseyPricey.js
```diff
@@ -0,0 +1,40 @@
+let peterHouse = {
```
LucasIversen: can be declared as const. Really nice approach with the object

javascript/javascript1/week1/homework_soheib/houseyPricey.js
```diff
@@ -0,0 +1,40 @@
...
+
+let juliaHouse = {
+    width: 5,
+    depth: 11,
+    height: 8,
+    garden: 70,
+    price: 1000000
+};
+
+function volumeCalc(house){
```
LucasIversen: didn't know you'd learned about functions yet! ;) Really nice approach

javascript/javascript1/week1/homework_soheib/houseyPricey.js
```diff
@@ -0,0 +1,40 @@
...
+    let volume = house.depth * house.width * house.height
+    return volume;
+}
+
+function houseValueCalc(house){
+    let houseValue = volumeCalc(house) * 2.5 * 1000 + house.garden * 300;
```
LucasIversen: all of these can be declared const, or you can also just return the calculation itself 👍

javascript/javascript1/week1/homework_soheib/houseyPricey.js
```diff
@@ -0,0 +1,40 @@
...
+    return houseValue
+}
+
+function compareValue(house) {
+    if(houseValueCalc(house) > house.price ) {
+        let diff = Math.floor(houseValueCalc(house)) - house.price; 
+        return console.log(`Ouch! This person is paying ${diff} too much for the house :(`)
+    } else if (houseValueCalc < house.price) {
```
LucasIversen: What is houseValueCalc without paramathers?

javascript/javascript1/week1/homework_soheib/scripts.js
```diff
@@ -0,0 +1,41 @@
+const birthYearInput = document.getElementById("birthYear");
+const futureYearInput = document.getElementById("futureYear");
+const targetInputContainer = document.querySelector('.ageIfy');
+const calculatebtn = document.getElementById("calculate");
+
+function ageIfy() {
```
LucasIversen, line 6: Daaaamn boy! This is nice 🔥

javascript/javascript1/week1/homework_soheib/scripts.js
```diff
@@ -0,0 +1,41 @@
...
+    }
+    const clearButton = document.getElementById("clear");
+    if (clearButton) {
+        clearButton.remove();
+    }
+    birthYearInput.value = "";
+    futureYearInput.value = "";
+}
+
+calculatebtn.addEventListener("click", ageIfy);
```
LucasIversen, line 41: My favorite part was line 1 to 41 ;)

LucasIversen (changes requested):
Gotta say, you really wen't for the gold Soheib! Just a few comments from me 👌

javascript/javascript1/week1/homework_soheib/ezNamey.js
siderdk: Thank you, I'll keep this perspective going forward

javascript/javascript1/week1/homework_soheib/goodboy_oldboy.js
siderdk: Oh I missed it, but for some reason it wasn't throwing an error, because I tested it on the browser and it wasn't running in strict mode. I asked about it before, and I was told that strict mode is by default now. However according to chatGPT that's not the case because it would break legacy code. 
I corrected it and tried to make the code shorter but still human-readable. 
Thanks a lot :D

javascript/javascript1/week1/homework_soheib/houseyPricey.js
siderdk: Thank you :D

javascript/javascript1/week1/homework_soheib/houseyPricey.js
siderdk: Sorrryyy :D 
I didn't think it was a big cheat, because we learned about function in the preparation materials anyway, and the practice goal of the exercise (if and comparisons) is inside the function. so two birds one logic :D

javascript/javascript1/week1/homework_soheib/houseyPricey.js
siderdk: 🫣

javascript/javascript1/week1/homework_soheib/houseyPricey.js
siderdk: That is a much better way. Many thanks

javascript/javascript1/week1/homework_soheib/houseyPricey.js
siderdk: I have rewritten it, so that it is more concise and readable. 
I stored the recurring calculations in values (const) and discovered Math.abs  that gives the absolute value, so that I don't need to calculate the difference twice. 
Many thanks for the very valuable insights :D

javascript/javascript1/week1/homework_soheib/scripts.js
siderdk: Thank you so much, I really appreciate it 🥲😄😄

LucasIversen (approved):

siderdk:
Making students feel like the mvp, is THE Real MVP 🙌🚀

PR #75, opened 2024-11-18
Javascript javascript1 week2/farzaneh

FarzanehAhmadi:


javascript/javascript1/week2/candy-helper.js
```diff
@@ -0,0 +1,62 @@
+let boughtCandyPrices = [];
+
+function addCandy(candyType , weight) {
+  if(candyType === 'Sweet')
```
julczka: The commented out solution is way better, just use that one. In general if you end up writing very long if-else statement, you should look for a better way of doing it, cause those are hard to read and to understand :) it also might be a good idea to do `candyType.toLowercase()` and compare with lower case candy names. But that is not a mistake by any means and it fits the requerements from the assignment.

javascript/javascript1/week2/candy-helper.js
```diff
@@ -0,0 +1,62 @@
...
+    boughtCandyPrices.push(weight*0.5);
+  
+  else if(candyType === 'Chocolate') 
+    boughtCandyPrices.push(weight*0.7);
+
+  else if(candyType === 'Toffee') 
+    boughtCandyPrices.push(weight*1.1);
+  
+  else if(candyType === 'Chewing-gum') 
+    boughtCandyPrices.push(weight*0.03);
```
julczka: It might be a better idea to simply return the candy value from this funcion, rather then pushing it to array from here. Ideally this code would be organized in more functions `checkCandyValue`, `buyCandy`, `canBuyMoreCandy` - then each funcion would do only one thing wich is usually how we want the code to work. It makes it easier to maintain and read.

javascript/javascript1/week2/candy-helper.js
```diff
@@ -0,0 +1,62 @@
...
+}
+
+const amountToSpend = Math.random() * 100;
```
julczka: this can be moved to the begining of the script

javascript/javascript1/week2/candy-helper.js
```diff
@@ -0,0 +1,62 @@
+let boughtCandyPrices = [];
```
julczka, line 1: That can be a `const`. Do you know why? Hint: array is a reference type

javascript/javascript1/week2/candy-helper.js
```diff
@@ -0,0 +1,62 @@
...
+  */
+  while ( i < boughtCandyPrices.length){
+    total += boughtCandyPrices[i];
+    i++;
+  }
+  return total;
+}
+
+
+let total = canBuyNewCandy(boughtCandyPrices);
```
julczka: This could be another function :)

javascript/javascript1/week2/candy-helper.js
```diff
@@ -0,0 +1,62 @@
...
+if (total === 0) 
+  console.log("You haven't bought any candy yet.");
+
+else if (total > amountToSpend)
+  console.log("Enough candy for you!");
+
+else console.log("You can buy more, so please do!");
+
+//This is another solution using the candyPrices object and cleaner code
```
julczka: So why this is commented out and the "ugly" one is the one you are using? 😉

javascript/javascript1/week2/event-application.js
```diff
@@ -0,0 +1,15 @@
...
+
+function getEventWeekday (daysToEvent){
+  const eventDayNumber = (dayNumber + daysToEvent) % 7;
+  return days[eventDayNumber];
+}
+
+console.log(getEventWeekday(5));
+console.log(getEventWeekday(9));
+
+//Here I used prompt for more practice :)
```
julczka, line 13: Very nice. You can do that, although I think no one is ever using prompt in real life :)

javascript/javascript1/week2/fullname-function.js
```diff
@@ -0,0 +1,17 @@
+function getFullname(firstname ,surname,  gender ='', useFormalName = true ){
+  
+  return firstname === '' || surname === '' ? 'Please provide both a first name and a surname.' :
```
julczka: Try to read this code in two weeks/a month and check if you can actually understand what it does :) 
As much as it is a nice excercise to use the ternary operator, this is an absolute nightmare to read and an instand headache :) Please avoid nesting more then two ternary operators in the future. Preferably use if statements with a return keyword following
```
if(someCondition) {
return someValue;
}
if(someOtherCondition) {
return someOtherValue;
}
//default case
return defaultValue;
```
Readability should always be a top priority when you write code, cause at work you will be returning to what you wrote multiple times and other people will also read and work on it too, so it must be understandable and not give anyone a headache 😉

javascript/javascript1/week2/student-manager.js
```diff
@@ -0,0 +1,25 @@
+const class07Students = [];
+
+function addStudentToClass(studentName) {
+   
+  if (studentName === 'Queen Margaret'|| studentName === 'Queen margaret' || studentName === 'queen margaret' || studentName === 'queen Margaret'){
```
julczka: Are you sure this is what requirements of this funtcion say? 
but assuming that it was that: is there any other way of checking that? Hint: string.split() and string.toLowerCase()

javascript/javascript1/week2/weather-wear.js
```diff
@@ -0,0 +1,13 @@
+function dressForTheWeather (temperature){
+  return temperature > -10 && temperature <= 10 ? 'Wear a Jacket, Hat and Gloves' :
```
julczka: Ouch, my head 🤯 😉

julczka (changes requested):
Hey, nice work, just avoid nesting multiple ternaty operators :)

FarzanehAhmadi:
Hi Julia,

Thank you so much for your feedback. It was really valuable and instructive! I’ll definitely keep my ternary operators on a leash from now on, no more headaches for anyone! 😄

Best,
Farzaneh

PR #77, opened 2024-11-19
Javascript-javascript1-week2/jalil

jalilhu:
fullname function
Event application
Weather wear
Student manager
Candy helper

### DIFF
--- a/javascript/javascript1/week2/candy_helper.js
+++ b/javascript/javascript1/week2/candy_helper.js
@@ -1,45 +1,40 @@
-const candyStore = 
-{   "Sweet": 0.5,
-    "Chocolate": 0.7, 
-    "Toffee": 1.1,
-    "Chewing-gum": 0.03
-}
-const boughtCandyPrices = []
+const candyStore = {
+  Sweet: 0.5,
+  Chocolate: 0.7,
+  Toffee: 1.1,
+  "Chewing-gum": 0.03,
+};
+const boughtCandyPrices = [];
 
-function addCandy(candyType, weight){
-    
-    let findWeight = candyStore[candyType]
-    const total = findWeight * weight
-    boughtCandyPrices.push(total);
-    return boughtCandyPrices; 
-    
-
+function addCandy(candyType, weight) {
+  let findWeight = candyStore[candyType];
+  const total = findWeight * weight;
+  boughtCandyPrices.push(total);
+  return boughtCandyPrices;
 }
 
 const amountToSpend = Math.random() * 100;
-console.log(amountToSpend)
+console.log(amountToSpend);
 
-console.log(addCandy("Sweet", 20))
-console.log(addCandy("Toffee", 10))
-console.log(addCandy("Chewing-gum", 750))
-console.log(addCandy("Chocolate", 30))
+console.log(addCandy("Sweet", 20));
+console.log(addCandy("Toffee", 10));
+console.log(addCandy("Chewing-gum", 750));
+console.log(addCandy("Chocolate", 30));
 
-function canBuyMoreCandy(){
-    let message = " "
-    let total = 0
-    for(let item in boughtCandyPrices){
-        total += boughtCandyPrices[item]
-        
-    }
-    console.log(total)
-    let CanIBuyMore = total ? total < amountToSpend : total > amountToSpend
-    if(total <= amountToSpend){
-        message = "You can buy more, please do"
-    }
-    else{
-        message = "Enough candy for you!"
-    }
+function canBuyMoreCandy() {
+  let message = " ";
+  let total = 0;
+  for (let item of boughtCandyPrices) {
+    total += item;
+  }
+  console.log(total);
+  let canIBuyMore = total ? total < amountToSpend : total > amountToSpend;
+  if (total <= amountToSpend) {
+    message = "You can buy more, please do";
+  } else {
+    message = "Enough candy for you!";
+  }
 
-    return CanIBuyMore + " " + message
+  return canIBuyMore + " " + message;
 }
-console.log(canBuyMoreCandy())
+console.log(canBuyMoreCandy());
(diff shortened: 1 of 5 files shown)


javascript/javascript1/week2/student_manager.js
```diff
@@ -0,0 +1,34 @@
+const class07Students = [];
+function addStudentToClass(studentName = String) {
```
NVedsted: This `= String` does not make a lot of sense here. It basically sets the default value to being the string "type". I'd recommend removing it. 🙂

(This explanation is overly simplified! If you want to read more about this, look up classes and constructors. To use an example you are familiar with: `Date` is a class and passing `Date` as a variable is like passing its constructor, i.e., the code called when you run `new Date()`, around. Don't worry if this is meaningless to you now; we will get to it later! 😄)

javascript/javascript1/week2/student_manager.js
```diff
@@ -0,0 +1,34 @@
...
+
+let message = "Only strings are allowed."
+    if(typeof studentName !=='string'){
+        return message;
+    }else if(studentName !== " " && !class07Students.find(name => name === studentName && getNumberOfStudents.length < 6)){
```
NVedsted: Good job on this implementation! I especially like your validation that a string was actually provided! ⭐ 

`.find` is a somewhat advanced feature, can you explain what it does in this case and what `=>` means here? 😄 Can you find another method for arrays that could simplify this? (Hint: jump over to [MDN's page on arrays](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array))

Something weird happens when you do `addStudentToClass("")` and nothing stops me from keeping to add people after the limit has been reached. Can you see why?  🐛 

There are also some duplicate code, i.e., line 7 shares some very similar code to line 13. Can you reorder your if-else-if-else such that this is not needed? This will likely also help with the bugs mentioned above! 😄

If you struggle with getting this to work, feel free to shoot me a message on Slack! 📬

javascript/javascript1/week2/event_application.js
```diff
@@ -0,0 +1,11 @@
+
+function getEventWeekDay(num){
+
+    const todaysDate = new Date();
+    const dayName = todaysDate.toLocaleString('en-US', {weekday: 'long'});
```
NVedsted: Nice and simple solution; excellent! 💪 Tiniest nitpick: is `dayName` needed? 😄

javascript/javascript1/week2/fullname.js
```diff
@@ -0,0 +1,9 @@
+function getFullName(firstname, surname, useFormalName=false, maleOrFemale=true){
+    const sex = maleOrFemale ? "Lord" : "Majesty";
```
NVedsted: I really like your structure here: nice and clean! 🤩 I would reconsider the variable name `maleOrFemale` as it is not clear what setting it to true or false means.

javascript/javascript1/week2/candy_helper.js
```diff
@@ -0,0 +1,45 @@
...
+
+console.log(addCandy("Sweet", 20))
+console.log(addCandy("Toffee", 10))
+console.log(addCandy("Chewing-gum", 750))
+console.log(addCandy("Chocolate", 30))
+
+function canBuyMoreCandy(){
+    let message = " "
+    let total = 0
+    for(let item in boughtCandyPrices){
```
NVedsted: Nice to not make your own indexes. Can you see how `for...of` could improve this? 🔍 

Remember to be consistent with your casing of variable names! `CanIBuyMore` starts with a capitalized C whereas others do not. For variables, I recommend sticking to lowercased initial letters. Many attribute a different meaning to variables starting with a capitalized letter as is evident with GitHub also showing it in a different color! 👀

javascript/javascript1/week2/weather_wear.js
```diff
@@ -0,0 +1,15 @@
+
+function suggestWhatToWear(num){
+    let youShoudWear = "nothing"
```
NVedsted: Nice and simple; good job! ⭐ Can you see how you can avoid this variable? 🔍 Something similar can be done in the student manager and candy helper! 😄

NVedsted (changes requested):
Nice work! 🌟 I have left a few comments with the main ones being on the student manager  which is also where I will recommend spending most of your follow-up time. 😄 

As a general, minor remark,  you have generally been very good with formatting (i.e., line breaks, indentation, and other spacing), but there are a few extra line breaks here and there. I would recommend using a formatter to make sure your code looks its best! Your editor can probably do it for you; give it a google if you are unsure how to! 💪 

javascript/javascript1/week2/student_manager.js
jalilhu: got it!
a little complicated 😄 
thanks.

javascript/javascript1/week2/student_manager.js
jalilhu: yes you are right
went through the code again. i found many mistakes
i added  1 more condition studentName === ""
i changed the whole code. 
thanks for the your review.

javascript/javascript1/week2/event_application.js
jalilhu: not really 😃 
i was googling and try to find out how to get the weekday
they were different options.

javascript/javascript1/week2/fullname.js
jalilhu: yeah you are right
always have problem with naming
now it is isItMale

javascript/javascript1/week2/candy_helper.js
jalilhu: yes easier and shorter 😺 
thanks for pointing it out.

i should check my code more often
there is formatting, variable names, for loops

javascript/javascript1/week2/weather_wear.js
jalilhu: instead of the variable i can return the string 
if it is 10 return "wear this"

PR #79, opened 2024-11-19
javascript-javascript1-week2/jianxin

jianxinz233:
Hand in homework of javascript1 week2

### DIFF
--- a/javascript/javascript1/week2/candyHelper.js
+++ b/javascript/javascript1/week2/candyHelper.js
@@ -1,0 +1,47 @@
+const boughtCandyPrices = [];
+
+function addCandy(candyType, weight){
+    const pricePerGram = {
+        sweet: 0.5,
+        chocolate: 0.7,
+        toffee: 1.1,
+        chewingGum: 0.03,
+    }
+    if(candyType === 'sweet'){
+       boughtCandyPrices.push(pricePerGram.sweet * weight);
+    }else if(candyType === 'chocolate'){
+        boughtCandyPrices.push(pricePerGram.chocolate * weight);
+    }else if(candyType === 'toffee'){
+        boughtCandyPrices.push(pricePerGram.toffee * weight);
+    }else if(candyType === 'chewingGum'){
+        boughtCandyPrices.push(pricePerGram.chewingGum * weight);
+    }
+}
+addCandy('sweet', 20);
+addCandy('chocolate', 30);
+addCandy('toffee', 10);
+addCandy('chewingGum', 50);
+console.log(boughtCandyPrices);
+
+let boughtCandySum = 0;
+// for loops:
+// for (let i = 0; i < boughtCandyPrices.length; i++) {
+//     boughtCandySum += boughtCandyPrices[i];
+// }
+
+// while loops:
+let i = 0
+while (i < boughtCandyPrices.length) {
+    boughtCandySum += boughtCandyPrices[i];
+    i++
+}
+console.log(boughtCandySum);
+
+
+let amountToSpend = Math.random() * 100;
+function canBuyMoreCandy() {
+    if(amountToSpend - boughtCandySum > 0){
+        return 'You can buy more, so please do!'
+    }else return 'Enough candy for you!';
+}
+console.log(canBuyMoreCandy());
--- a/javascript/javascript1/week2/eventApplication.js
+++ b/javascript/javascript1/week2/eventApplication.js
@@ -1,0 +1,10 @@
+const weekDays = ['Sunday', 'Monday', 'Tuesday', 'Wednesday', 'Thursday', 'Friday', 'Saturday'];
+const todaysDay = new Date();
+
+function getEventWeekday(days) {
+    const eventDayNum = (todaysDay.getDay() + days) % 7;
+    return weekDays[eventDayNum];
+}
+
+console.log(getEventWeekday(2));
+console.log(getEventWeekday(10));
--- a/javascript/javascript1/week2/flightFullname.js
+++ b/javascript/javascript1/week2/flightFullname.js
@@ -1,0 +1,32 @@
+// function getFullname(firstname, surname){
+//     return firstname + ' ' + surname;
+// }
+
+
+// // Formal fullname
+// function getFullname(firstname, surname, useFormalName) {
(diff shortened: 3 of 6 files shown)


javascript/javascript1/week2/flightFullname.js
```diff
@@ -0,0 +1,32 @@
...
+
+// // Formal fullname
+// function getFullname(firstname, surname, useFormalName) {
+//     return useFormalName ? 'Lord ' + firstname + ' ' + surname : 
+//     firstname + ' ' + surname;
+// }
+
+
+function getFullname(firstname, surname, useFormalName=true, gender) {
+    return firstname === '' || surname === '' ? 'Please provide your firstname and surname.' :
```
NVedsted, line 14: Ternary operators can be cool, but it is so easy to get lost in them. 🥴 I would highly recommend to never nest them inside each other and instead use if-elseif-else. 🪆 This is mainly because it is not intuitive what belongs to what. You can take a look [here](https://stackoverflow.com/questions/73603909/right-associativity-and-order-of-execution-of-nested-ternary-operator-in-c) for someone else who tried to decipher nested ternaries. 😄

Also a good opportunity to try some [template strings](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals)! 🤩

javascript/javascript1/week2/flightFullname.js
```diff
@@ -0,0 +1,32 @@
...
+
+function getFullname(firstname, surname, useFormalName=true, gender) {
+    return firstname === '' || surname === '' ? 'Please provide your firstname and surname.' : 
+    useFormalName && gender === 'male' ? 'Lord ' + firstname + ' ' + surname : 
+    useFormalName && gender === 'female' ? 'Lady ' + firstname + ' ' + surname :
+    firstname + ' ' + surname;
+}
+
+
+const fullname1 = getFullname('Alice', 'White', 'female');
```
NVedsted, line 21: Does `getFullname('Alice', 'White', 'female')` return what you expect? And if so or not so, why? 🧐

javascript/javascript1/week2/studentManager.js
```diff
@@ -0,0 +1,29 @@
+const class07Students = [];
+function addStudentToClass(studentName) {
+    if(studentName === ''){
```
NVedsted, line 3: Excellent chain of conditions! Good job! ⭐

javascript/javascript1/week2/eventApplication.js
```diff
@@ -0,0 +1,10 @@
+const weekDays = ['Sunday', 'Monday', 'Tuesday', 'Wednesday', 'Thursday', 'Friday', 'Saturday'];
+const todaysDay = new Date();
+
+function getEventWeekday(days) {
```
NVedsted, line 4: Nice and simple! 🤩 What would happen if I opened your page just before midnight and started using it after midnight? 🤔 Can you see what unexpected behavior might occur? And if so, how would you mitigate it? 😄

javascript/javascript1/week2/candyHelper.js
```diff
@@ -0,0 +1,47 @@
...
+
+let boughtCandySum = 0;
+// for loops:
+// for (let i = 0; i < boughtCandyPrices.length; i++) {
+//     boughtCandySum += boughtCandyPrices[i];
+// }
+
+// while loops:
+let i = 0
+while (i < boughtCandyPrices.length) {
```
NVedsted, line 34: `while` loops get the job done! 💪 But what would happen had you forgotten to increment `i` inside it? Generally, when going over arrays, I recommend using [for...of](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...of) as it is less error-prone. 😄 

If I were to call `canBuyMoreCandy` and was told that I can, and I then spend all my money. What would `canBuyMoreCandy` return afterwards? Can you see what might cause issues here? 🐛

NVedsted (changes requested):
Very neatly structured and pleasant-to-read code; good job! 🤩 I have added some comments that I'd like you to take a look at it. 🔍 

Let me know if you have any questions or issues. 😄

javascript/javascript1/week2/flightFullname.js
jianxinz233: Thank you for the reviews!
Actually I also feel if-elseif-else is more clear, but I'm not so familiar with  ternary operator, so I just want to have a try. 😂 And the template strings, I will work on it, Thanks! 
And I will update and improve the homework later.

javascript/javascript1/week2/flightFullname.js
jianxinz233: Here I want to show the case if passenger didn't choose if she need a formal name, not providing information, then it return as normal full name, though she provided gender. Here it seems to work...? 🤔

javascript/javascript1/week2/eventApplication.js
jianxinz233: I think I can understand the situation, but sorry, I don't know the answer... Let me have a try.

javascript/javascript1/week2/candyHelper.js
jianxinz233: I also prefer for loop, as I always found it ran errors then found that "i" is missing 😅 But as an optional homework of while loop, I put both here.
About the second question, I will try to figure it out.
Thank you for all the reviews! After finishing this week's homework, I will update the changes. Have a nice evening! ❤️

javascript/javascript1/week2/flightFullname.js
NVedsted: But your default value for `useFormalName` is true? 😄 I will just cut to the chase, haha, in this case you are passing the value `'female'` as an argument into `useFormalName`, but since you are not passing a `gender`, it does not try to add Lord or Lady. I hope it makes sense. 🙂

javascript/javascript1/week2/eventApplication.js
NVedsted: It is a bit convoluted to spot the issue here, but imagine you had this script loaded onto a website and a user pressing a button would run `getEventWeekday(...)` and the user did not close the window for multiple days. What happens with `todaysDay` then? 😄

PR #81, opened 2024-11-20
Javascript javascript1 week2/ruslana

ruslanaonshchk:
added solving mandatory tasks

### DIFF
--- a/javascript/javascript1/week2/script.js
+++ b/javascript/javascript1/week2/script.js
@@ -116,3 +116,58 @@ addStudentToClass('Queen');
 addStudentToClass('');
 
 console.log(`Total students in the class: ${getNumberOfStudents()}`);
+
+// OPTIONAL TASK : Candy helper
+
+const boughtCandyPrices = [];
+
+function addCandy(candyType, weight) {
+  const candyPrices = {
+    sweet: 0.5,
+    chocolate: 0.7,
+    toffee: 1.1,
+    'chewing-gum': 0.03,
+  };
+
+  if (candyType in candyPrices) {
+    const price = candyPrices[candyType] * weight;
+    boughtCandyPrices.push(price);
+  } else {
+    console.log('Invalid candy type!');
+  }
+}
+
+const amountToSpend = Math.random() * 100;
+
+function canBuyMoreCandy() {
+  let totalPrice = 0;
+
+  for (let price of boughtCandyPrices) {
+    totalPrice += price;
+  }
+
+  // using while loop
+  /*
+    let i = 0;
+    while (i < boughtCandyPrices.length) {
+        totalPrice += boughtCandyPrices[i];
+        i++;
+    }
+    */
+
+  if (totalPrice < amountToSpend) {
+    console.log('You can buy more!');
+    return true;
+  } else {
+    console.log("That's enough candy for you!");
+    return false;
+  }
+}
+
+addCandy('sweet', 20);
+addCandy('chocolate', 15);
+addCandy('toffee', 10);
+addCandy('chewing-gum', 50);
+console.log('Bought Candy Prices:', boughtCandyPrices);
+
+canBuyMoreCandy();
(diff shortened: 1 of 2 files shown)


javascript/javascript1/week2/script.js
```diff
@@ -0,0 +1,173 @@
...
+    console.log('Invalid candy type!');
+  }
+}
+
+const amountToSpend = Math.random() * 100;
+
+function canBuyMoreCandy() {
+  let totalPrice = 0;
+
+  for (let price of boughtCandyPrices) {
```
NVedsted, line 145: Excellent way of summing up the prices! ⭐ Can you explain why you decided to go with this approach over the `while` loop below? 😄

javascript/javascript1/week2/script.js
```diff
@@ -0,0 +1,173 @@
...
+}
+
+const clothesToWear = whatToWear(22);
+console.log(clothesToWear);
+
+// TASK : Student manager
+
+const class07Students = [];
+function addStudentToClass(studentName) {
+  if (!studentName) {
```
NVedsted, line 84: Easy to follow and great use of early-exits! 🌟

javascript/javascript1/week2/script.js
```diff
@@ -0,0 +1,173 @@
+// TASK Step 3: Smart-ease - Goes Global! : Flight booking fullname function
+
+function getFullname(
+  firstname,
+  surname,
+  useFormalName = false,
+  gender = 'male',
+) {
+  if (!firstname || !surname) {
+    return 'Invalid name';
```
NVedsted, line 10: Nice to see early-exiting! 👍 Can you explain why this is preferrable to putting the rest of the method in an `else`? 😄 

Furthermore, can you see where more early-exiting can be applied further down the method? 🔍

javascript/javascript1/week2/script.js
```diff
@@ -0,0 +1,173 @@
...
+  return weekdays[eventDayIndex];
+}
+
+console.log(getEventWeekday(9));
+console.log(getEventWeekday(3));
+
+// TASK : Weather wear
+
+function whatToWear(temp) {
+  if (temp >= 25) {
```
NVedsted, line 64: Very clean use of `else-if` chains; awesome! 💪

NVedsted (approved):
Excellent submission! 🌟 I did not spot any bugs, the code was formatted neatly, and the logic was easy to follow; keep up the awesome work! 🤩 

I have added some questions to your choices that I'd love to hear your replies to! 😄 

Let me know if you have any questions or issues.

javascript/javascript1/week2/script.js
ruslanaonshchk: thank you so much for such good comments and questions; 
so for me, **for... of** statement looks clearer and takes fewer lines

javascript/javascript1/week2/script.js
ruslanaonshchk: I'm not sure here, but I think, we could return the formal name immediately if useFormalName is true, avoiding extra conditions.

NVedsted: What if we flipped it around, i.e., if it is not a formal name?

ruslanaonshchk: sorry, I forgot to answer 
I'm not sure, that I got your question

NVedsted: No worries! 😄 What I was fishing for here is something like this:
```js
if (!useFormalName) {
  return firstname + ' ' + surname;
}

if (gender === 'male') {
  return 'Lord ' + firstname + ' ' + surname;
} else if (gender === 'female') {
  return 'Lady ' + firstname + ' ' + surname;
} else {
  return 'Dear ' + firstname + ' ' + surname;
}
```

By flipping the first condition, we can exit early and avoid wrapping the rest in an `else`. This reduces indentation and makes the code a bit easier to to follow. 🙂

PR #82, opened 2024-11-20
Javascript javascript1 week2/soheib

siderdk:
The optional homework is still in progress, I will write when it's done. 
Thank you in advance for you help.


javascript/javascript1/week2/homework4/eventApp.js
```diff
@@ -0,0 +1,13 @@
+const weekDays = ['Monday', 'Tuesday', 'Wednesday', 'Thursday', 'Friday', 'Saturday', 'Sunday'];
+const today = new Date()
+function getEventWeekDay(days) {
+    const day = today.getDay();
+    let result;
+    if(day + days <7){
+        result = day + days -1;
```
NVedsted: You got the right idea with the remainder operation; awesome! ⭐ Can you think of a to avoid the `if` statement? (Hint: what is `(day + days - 1) % 7` equals to when the sum is below 7?)

Another minor thing is the `-1`; can you think of a way to avoid it? (Hint: what's the first day of the week according to `getDay`?) 😄

javascript/javascript1/week2/homework4/smartEase.js
```diff
@@ -0,0 +1,29 @@
...
+const fullName1 = getFullnameV1('Mark', 'Corrigan');
+const fullName2 = getFullnameV1('Jeremy', 'Usbourne');
+console.log(fullName1);
+console.log(fullName2);
+
+function getFullnameV2(firstName, surname, useFormalName=true, gender) {
+    if((useFormalName === undefined || useFormalName === false) && (gender === undefined || gender === 'other')){
+        return getFullnameV1(firstName, surname)
+    }
+    else if(useFormalName === true){
```
NVedsted: As a general rule of thumb, I'd suggest avoiding comparing with booleans as you can simple do `if (useFormalName)` and it behaves identically! 😄 Can you think of how we'd check for something being false without using a comparison? (Hint: you will need an operator!)

javascript/javascript1/week2/homework4/smartEase.js
```diff
@@ -0,0 +1,29 @@
...
+  return fullName;
+}
+
+const fullName1 = getFullnameV1('Mark', 'Corrigan');
+const fullName2 = getFullnameV1('Jeremy', 'Usbourne');
+console.log(fullName1);
+console.log(fullName2);
+
+function getFullnameV2(firstName, surname, useFormalName=true, gender) {
+    if((useFormalName === undefined || useFormalName === false) && (gender === undefined || gender === 'other')){
```
NVedsted: Nice reuse of your first implementation! 👍 If we were to assume the old one would no longer be used directly, I would probably avoid having it as a separate function, but we do not know that here. 😄

As a small nitpick, can you think of a way to avoid the `fullFormalName` and `fullName` variables? 🔍

javascript/javascript1/week2/homework4/studentManager.js
```diff
@@ -0,0 +1,43 @@
...
+function addStudentToClass(studentName) {
+    if (studentName === '' || studentName === undefined) {
+        return `A student name is missing.`;
+      }
+    
+      if (class07Students.includes(studentName)) {
+        return `${studentName} is already in the class.`;
+      }
+    
+      if (class07Students.length >= 7 && studentName !== 'Margarethe') {
```
NVedsted: Very nice and straightforward implementation! 🤩 I do think our other Queen, Mary, would be bummed out if she tried to join a full class though. 🥺😛

javascript/javascript1/week2/homework4/weatherClothApp.js
```diff
@@ -0,0 +1,75 @@
...
+    for (let key in obj) {
+        const valuesArray = obj[key];
+        const randomIndex = Math.floor(Math.random() * valuesArray.length);
+        result[key] = valuesArray[randomIndex];
+    }
+    return result;
+}
+
+function getClothes(degrees){
+    if (degrees === undefined) { return `please enter and int value of the temperature in Celsius`}
```
NVedsted, line 50: I have never seen anyone go into this amount of detail on this task, and I love it! 🤩 Excellent use of template strings as well. 😄 

As there are some complex features and tricks in use, I just wanna hear you explain some of the concepts in your own word:
 - What does a `for..in` loop do? And how does it relate to arrays?
 - What does `clothes.coats || 'None'` do and how does it work?

Should you get bored, I think this function would benefit greatly from another helper function that takes a `clothes` object and pretty-prints based on the properties. In that way, you could shorten it a lot! **I do not expect you to do this**, but I think it would be a good exercise to push your skills further if you want to. 😄

NVedsted (approved):
Excellent work that touches a lot of JavaScript concepts and features! 🤩 I have put some comments on your code that I would like you to address. 😄 

Your weather clothes solution was a lot of fun to review! I have put a suggestion for improvement on it that would truly utilize your existing code that you can have some fun with if you have the time and want to. 👍 

javascript/javascript1/week2/homework4/eventApp.js
siderdk: Thank you so much for the very pedagogical advice. 
I do have a tendency to over-complicate my life 😄
I have reordered the array to remove the -1, and simplified the logic to remove the unnecessary if. 
I also made the today const local to the function chained the getday to new date. 
A lot cleaner and more readable 🙏

javascript/javascript1/week2/homework4/smartEase.js
siderdk: How did I miss that !!!
I cleaned the code, introduced a fallback value for gender
Thank you so much 🙏🙏

javascript/javascript1/week2/homework4/smartEase.js
siderdk: I did get rid of the call to v1 from inside v2, which I falsely mentioned as "nesting" in my commit message. 
I also got rid of the variables by introducing a prefix. and now it does look a lot cleaner and more concise 🙏
Can you please confirm this is the goal of your comment, otherwise I will be more than happy to tidy up more, if it is possible.

javascript/javascript1/week2/homework4/studentManager.js
siderdk: I added her Highness Queen Mary to the exception rule alongside the Queen-mother Margarethe II. May the monarchs and the royalists be happy now 😄😄

javascript/javascript1/week2/homework4/smartEase.js
NVedsted: This is a nice change! 👍 A smal thing is that maybe the gender check should be OR'd instead of AND'd. If I read this code correctly, I would expect it to yield `" Firstname Lastname"` (note the whitepace in the beginning) if you ask for a formal name but use the gender `"other"` which is a bit unexpected.

As a minor formatting thing, I would also avoid nesting ternaries as they quickly get out of hand and become hard to read. 😄

siderdk: I have fixed it. I didn't pick up on the unintended gender assignment, which is to think of it, the opposite of the initial goal. Thank you for that! 
I also for rid of the nested ternaries for a more readable if else. 
Thank you so much.

javascript/javascript1/week2/homework4/weatherClothApp.js
siderdk: This was really instructive. I am really grateful for the very instructive and illuminating advice. I have added the function you suggested, and it did make this a lot more readable, and that was not because I was bored in any way 😄😄
To answer your question: 

- The `for..in` iterates through a certain interval that we determine, while performing a task. in an array this specific syntax will iterate through the indices of the said array.  However I have just read through some documentation and got reminded that: `for..in` does not only iterate through the content of the array, but also through properties that are not necessarily indexed or enumerated. So if I understand correctly, I should better use `for...of` instead, right? 
- `clothes.coats || 'None'` checks for left value first, and if it's falsy it will check the one on the right. I used this as I was still populating the objects, and I didn't want an `undefined` value showing up. As things stand, I don't need it, especially since now there's a function to "pretty-print" that part. 

This was a lot of fun, I am very grateful.

NVedsted: Regarding `for...in` and `for...of`, put very simply the former loops through properties and the latter through elements. If you do `in` on an array, it spits out the indexes (and any properties that it might have as well) which is a bit error-prone. If you need to iterate through arrays, `of` is definitely the safer choice as it just yields the values. 😄 

It also sounds like you kinda encountered a gotcha with `in`: that it can pick up on weird, unexpected properties. You might have seen `hasOwnProperty` already which is generally how you'd avoid this. But yea, we are getting way ahead of ourselves here. 😆 

Good works and thanks for humoring me! 🤩

siderdk: Best deal ever, I get knowledge, you get laughs 😄 
I am very grateful. This was really insightful.

PR #83, opened 2024-11-21
javascript-javascript1-week2/hossein

hossein1065:


javascript/javascript1/week2/Flight_booking_fullname_function.html
```diff
@@ -0,0 +1,59 @@
...
+    getFullname(fName,lName,titleLord);
+
+      ">Print my ticket</button>
+  </div>
+  <div>
+    <textarea name="" id="fullname"  readonly> </textarea>
+  </div>
+  
+  <script>
+    function getFullname(firstname,surName,addLord){
```
NVedsted, line 48: Very nice showcase of the functionality! ⭐ The homework also asked you to take gender into account, how would you do that? 😄 

You are using [template strings](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals) extensively below, but not in their intended way. Can you see what you could do differently? 🔍

NVedsted, line 48: Very nice showcase of the functionality! ⭐ The homework also asked you to take gender into account, how would you do that? 😄 

You are using [template strings](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals) extensively below, but not in their intended way. Can you see what you could do differently? 🔍

javascript/javascript1/week2/other_task.js
```diff
@@ -0,0 +1,176 @@
+
+
+// Event application
+function getEventWeekday(day){
+  const weekDays=['Sunday', 'Monday', 'Tuesday', 'Wednesday', 'Thursday', 'Friday', 'Saturday'];
+  const currentDate  = new Date();
+  const dayNumbers = currentDate.getDay();
+   const futureDay= (day + dayNumbers) % 7;
+   console.log(weekDays[futureDay]);
```
NVedsted, line 9: Excellent use of remainder! 😄 Developers using your function might be a bit surprised that a function called `getX` does not return anything; maybe it should be up to the caller to decide whether to log it or not? 📞

javascript/javascript1/week2/other_task.js
```diff
@@ -0,0 +1,176 @@
...
+}
+getEventWeekday(1);
+
+
+// Weather wear 
+ function SuitableClothes(temperture){
```
NVedsted, line 15: Good use of if-else-if-else! ⭐ The function name can be misleading though! Many will likely expect something else when seeing the starting letter capitalized; you can even see that GitHub renders it differently. 👀

This is all convention though as has no functional impact in JavaScript. My suggestion is to stick to standard conventions or at least be consistent within your project, so to stick with convention and be consistent with your other functions, I'd call this function `suitableClothes`. 😄 

I would not recommend your alternative `switch`-based approach as it is a lot more error prone and generally harder to maintain due to the weird, different syntax. 🥴

javascript/javascript1/week2/other_task.js
```diff
@@ -0,0 +1,176 @@
...
+    return;
+  }
+  
+  if (studentName===``){
+    console.log(`Please put valid name`);
+    return;
+  }
+  
+  
+  if(class07Students.length>=5 ){
```
NVedsted, line 84: Nice structure of the function with early-exiting and such! 🤩 However, it does look like people are being let into the class even if they are told there's no room! 😮 Can you spot the bug? 🐛 (Hint: what happens when `name7` is added below?)

I believe the intended behavior of `getNumberOfStudents` is to simply get the total number of students in the class, but what you have implemented is also interesting! 😄

javascript/javascript1/week2/other_task.js
```diff
@@ -0,0 +1,176 @@
...
+        console.log(`Enough candy for you!`);
+      }
+    
+
+    
+  }
+   function sumPrices(){
+    let sum=0;
+    let i=0;
+    while
```
NVedsted, line 158: I would recommend generally staying clear of `while` when you can as they are error-prone; e.g., what would happen if you forgot to increment `i`? 🤯 

In this case, and generally when iterating over arrays, I would recommend [`for...of` loops](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...of). 😄

NVedsted (changes requested):
Very nice work! ⭐ I really liked the HTML form you made for showing off your implementation; I did not review the non-JS part as it is beyond the scope of this hand-in though.

I would recommend spending more time on formatting of your code in the future as both you and your future colleagues will be a lot happier if your code is easier to read and navigate. 😄 Your editor likely supports this, so google around to see what is possible. 🔍 

Feel free to reach out if you have any questions! 😃

hossein1065:
thanks for reviewing my task and giving me advice I will try to follow your advice on my code

PR #84, opened 2024-11-21
Javascript javascript1 week2/andriiK

pahanan:


pahanan:
the first time I accidentally added files was in the first week.

javascript/javascript1/week1/flight_booking.js
```diff
@@ -0,0 +1,52 @@
+function getFullname(firstname, surname, useFormalName = false){   // I added false as the default boolean value, if we don't have false or true as an argument
```
julczka: That is very nice case for default parameter value! good job!

javascript/javascript1/week1/flight_booking.js
```diff
@@ -0,0 +1,52 @@
...
+    if(typeof(firstname) === "string" && typeof(surname) === "string"){
```
julczka: Great that you arfe checking for the type of a value passed to the funcion... I see one trap here though.
Try and run this line in the console `typeof ""`. What are you getting? Also, remember that `"1234"` and `1234` are of a different type. Try to test your funcion [on this page](https://codepen.io/hackyourfuture-cph/pen/jJWwbN) and see waht is happening when you type in numbers to the inputs :)

Aditional "bonus" check could be to make sure we are actually passing some letters in there, like characters from A to Z etc. You usually do it with something called regex. You dont have to worry about that now, as it is notoriuosly tricky, just google that if you are curious :)

javascript/javascript1/week1/flight_booking.js
```diff
@@ -0,0 +1,52 @@
...
+        if(useFormalName){
+            return ("Lord " + firstname + " " + surname);
+        } else{
+            return (firstname + " " + surname);
+        }
+    } else{
+        return ("Firstname or surname are not correct!");
```
julczka: On this line, instead of returning a string, you could use `throw` expression. [You can read more about it here.](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/throw)
This is also might be bit advanced for now, but you can try to replace this line with `throw "Firstname or surname are not correct!"` ans see in the console what it says there.

javascript/javascript1/week1/flight_booking.js
```diff
@@ -0,0 +1,52 @@
...
+
+console.log(fullname1);
+console.log(fullname2);
+console.log(fullname3);
+
+
+// Works for women
+
+
+// function getFullname(firstname, surname, isMale, useFormalName = false ){
```
julczka: I think in real life the best approach would be to let user choose their title themselves :) That is just a general remark, you do not have to fix that of course :)

javascript/javascript1/week1/event_application.js
```diff
@@ -0,0 +1,19 @@
...
+            numberDay--;
+            continue;
+        }
+        i++;
+        numberDay--;
+    }
+    return("Event will be on " + days[i]);
+}
+
+const today = new Date();
```
julczka: I would define these two variable inside `getEventWeekday` funtcion. I am not sure if you had that before, but all variables have something called scope. If you define something in your script, you can access it in any function in this script. But if you define it inside the funcion, you wont be able to access that variable outside of this function. You can try to do that by moving this to lines to line 2 and try to console log them on line 19 - this will result in an error. 
It would make sense to define them in the funcion scope because this is where we need them - if by accident some other function in this script would reassign the value of `days` then `getEventWeekday` would stop working. Also the `new Date()` will only run once and not every time we run this funcion. So, if we use this today, leave the browser window open for 24 hours and use the funcion again I think the result will be incorrect baceuse we will still use the value from yesterday. if you define today inside your funcion then each time you call the function that will get a new value.

javascript/javascript1/week1/event_application.js
```diff
@@ -0,0 +1,19 @@
+function getEventWeekday(numberDay){
+    let i = today.getDay();
```
julczka: As much as this funcion works (or at least from what I tested, I'm not sure it it is not going to an infinite loop when you put 0 as an argument :) ) it is bit hard to understand what is going on here. You can try to rewrite it with the use of remainder operator. Try typing in the console `3%7` then `14%7` then `23%7` and see what you get. Do you see any pattern? Now, if our funcion would only be run on Sundays that would be quite easy, but it is not. So we need to offset the output of remained operation by the number for days that already passed since sunday, right? Give it a try if you want to :)

javascript/javascript1/week1/student_manager.js
```diff
@@ -0,0 +1,58 @@
+const class07Students = [];
+function addStudentToClass(studentName) {
+    if(typeof(studentName)==="undefined" || studentName.length == 0){
```
julczka: You can simplify that a bit by writing (always use tripple equals, the second condition has a double one)
```
if(studentName === undefined || studentName === '') 
```
in case of the string even something like this will work (although some programmers do not like that way of checking for empty value)
```
if(!studentName)
```
that works because of the type coversion in javascript - it will try to change the type of whatever `studentName` is to Boolean, so if it is empty string, `null`, `undefined` or `0` it will be changed into `false`. So we check for `if(!true)` or `if(!false)` - you can check what they evaluate to by typing `const x = !true` in the console and check what the value is, then you will know when what your if statement will evaluate too. Both of those ways have theis pros and cons I guess. First one is more explicit, the other would hadle more values. I guess the safest way would be to check for `typeof studentName !== 'string' || studentName === ''`

javascript/javascript1/week1/student_manager.js
```diff
@@ -0,0 +1,58 @@
...
+        return(console.log("Name is not correct!"));
+    }
+
+    for (let i = 0; i < getNumberOfStudents(); i++) {
+        if(class07Students[i]==studentName){
+            return(console.log(`Student ${studentName} is already in the class.`));
```
julczka: You do not need the return statement before the call to console.log. You can just console.log things this code path will return `undefined` anyway. That is the default return value of a funcion when there is nothing to return :) You can check the return value of `console.log("Some message")`. Try to store it in a variable and the console log this variable. What do you get?

javascript/javascript1/week1/student_manager.js
```diff
@@ -0,0 +1,58 @@
...
+        }
+    }
+
+    if(getNumberOfStudents() == 6 && studentName!=="Queen"){
```
julczka: Always use === :)

javascript/javascript1/week1/student_manager.js
```diff
@@ -0,0 +1,58 @@
...
+        return(console.log("Cannot add more students to class 07"));
+    }
+
+    if(studentName==="Queen"){
+        if(getNumberOfStudents() == 6){
+            class07Students.shift();
```
julczka: I'm not sure if you need all this. I think queen will just be added to the end array with push, no probelm. I did not test, but my gut feeling tells me that if you remove this entire block and test again it will work more less the same, except you wont kick anyone out from the class as you do now :) But that all depends on how we understang the requerements - I read them as queen can be 7th member of the class and you stick to 6 as max number. It is an interesting case of how requirements are not always super clear :) very much like real life :)

javascript/javascript1/week1/student_manager.js
```diff
@@ -0,0 +1,58 @@
...
+            class07Students.unshift("Queen");
+            return (console.log("Queen added"));
+        } else {
+            class07Students.unshift("Queen");
+            return (console.log("Queen added"));
+        }
+    }
+
+    if(typeof(studentName)==="string"){
```
julczka: Try to think about requirements and edge cases in this funcion and the order that we should check them. We know that:

1. This funcion only accepts strings
2. String cannot be empty
3. Class cannot be bigger then 6
4. We cannot add the same person
5. We have to handle "Queen" in a special way
 Try to order those in a funcion from the "broadest/most important" to the narrowest and then the default behavior. It improves readability.

javascript/javascript1/week1/weather_wear.js
```diff
@@ -0,0 +1,20 @@
+function whatWearToday(temperature){
+    if(temperature <= 5 && temperature > -20){
```
julczka: You can try and order those if statements by temperature, so the coldest and hottest will be at the begining and end of the funcion. It will make it more readable. In general - this is a very good way of writing: if(some condition) return something instead of writing if-elseif-else so stick with it :)

javascript/javascript1/week1/weather_wear.js
```diff
@@ -0,0 +1,20 @@
...
+        return("Put on warm pants, a warm jacket and a hat.");
+    }
+    if(temperature<= -20){
+        return("Where are you??? Stay home!");
+    }
+    if(temperature > 5 && temperature <= 18){
+        return("Put on your pants and a sweatshirt.");
```
julczka: You dont need parenthesis around the return value. you can just write 
`return "Some message"`

javascript/javascript1/week2/candy_helper.js
```diff
@@ -0,0 +1,57 @@
...
+    if (totalPrice < amountToSpend) {
+        console.log("You can buy more, so please do!");
+        return true;
+    } else {
+        console.log("Enough candy for you!");
+        return false;
+    }
+}
+
+let boughtCandyPrices = [];
```
julczka: I would move that to the begining of the script. It technically works when defined here but it's hard to read he funcions that are using that one before you can see the definition.

javascript/javascript1/week2/candy_helper.js
```diff
@@ -0,0 +1,57 @@
+function addCandy(candyType, weight) {
+    let pricePerGram = 0;;
+
+    switch (candyType.toLowerCase()) {
+        case "sweet":
```
julczka, line 16: if you want to be super organized, you can store those candy kind strings in variables at the beging of the script.
```
const TOFFE = "toffee"
```
and then use that in switch case 
```
case TOFFE: 
```
that would be a good practice that will allow to change such value easily in the future it all of a sudden "toffee" is callled something else and used in many places in this script. It is a convention to write the names of such variables upper case - these are values thatg are predefined before runtime. Other example would be max number of children in a class from the  student array excercise.

julczka: And if you want to be very fancy, you could also organize all these in an object, for example 
```
const candyMenu = {
toffee: {
  typeName: 'toffe',
  price: 1.1
 }
//add more candy here
}
```
then you can access particular values through `candyMenu.toffee.price` and if the type is dynamic you can even do `candyMenu[candyType].price`
That would be an "advanced" way of doing that, I'm just showing this as an example of how you can organize data in your code. If that is not clear then come back to this comment in a week after you had objects :)

julczka (changes requested):
Very nice work! You have duplicated files in this PR, you can just remove the duplicates, so there is no mess. 
I like very much how you write your funcions, it looks like you know what you are doing, so I left you some comments mostly about how to improve readability or how to do something in a different way. If any of those are unclear let me know below or find me on slack :) my full name is julia gruszczynska

pahanan:
Thank you for your detailed comments. It was interesting to read this and learn something. After reading the letter in the mail, I started deleting files from the first week, and later I saw that you left commits for the first week. So I'm already confused about what to delete :))

PR #86, opened 2024-11-21
Javascript javascript1 week2/yuusuf

Ganja0003:


javascript/javascript1/week2/event-app.js
yousseftopaji, line 1: Very good, and clever choice to find what the day is through the method Date().getDay().

javascript/javascript1/week2/flight-booking.js
```diff
@@ -0,0 +1,18 @@
...
+
+function getFullname(firstname,surname,useFormalName) {
+    if (useFormalName == true){
+        return ("lord " + firstname + surname)
+    }else{
+        return(firstname + surname);
+    }
+
+    
+}
```
yousseftopaji, line 11: Great work!
I have a challenge for you: if the customer is female and she decides to be called in a formal way, she will be called Lord.
What would you do to fix this?

javascript/javascript1/week2/student-manager.js
```diff
@@ -0,0 +1,41 @@
+const class07Students = [];
+function addStudentToClass(studentName) {
+  
+    studentName === "" ? console.log("Please add a name!"):undefined;
```
yousseftopaji, line 4: The ternary operator is used here but the undefined part is unnecessary. It would be clearer to use an if statement.
And if you try to add an empty name, it will still be added.
So it is best to make a consistent if, else if kind of statements.
```suggestion
if (studentName === "") {
    console.log("Please add a name!");
    return;
  }```

File: javascript/javascript1/week2/student-manager.js [hunk 3]
```diff
@@ -0,0 +1,41 @@
...
+
+    if (studentName == "Queen"  || studentName === "queen"){
```
yousseftopaji, line 6: - I added "else" to follow up on the previous comment, so it is more consistent where it will no longer approve adding an empty name

- Use strict equality (===) for both comparisons in the condition if (studentName == "Queen" || studentName === "queen") for consistency.

- and you can simply use toLowerCase() method.


```suggestion
      else if (studentName.toLowerCase() === "queen") {
```

javascript/javascript1/week2/student-manager.js
```diff
@@ -0,0 +1,41 @@
...
+        class07Students.push("Queen");
+        return;
+    } else if(class07Students.length >= 6 && studentName !== "Queen"){
```
yousseftopaji, line 9: Great usage of JS local scope.
You have already passed the condition (if student name === queen => push to the array), so you don't need to mention it again.

javascript/javascript1/week2/student-manager.js
```diff
@@ -0,0 +1,41 @@
...
+        return "Sorry there is no more space for other students.";
+    }  else if (class07Students.length < 6 && !class07Students.includes(studentName) ) {
```
yousseftopaji, line 11: same as before, you don't need this part of the statement "class07Students.length < 6 "

javascript/javascript1/week2/student-manager.js
```diff
@@ -0,0 +1,41 @@
+const class07Students = [];
+function addStudentToClass(studentName) {
+  
+    studentName === "" ? console.log("Please add a name!"):undefined;
+
+    if (studentName == "Queen"  || studentName === "queen"){
+        class07Students.push("Queen");
+        return;
+    } else if(class07Students.length >= 6 && studentName !== "Queen"){
+        return "Sorry there is no more space for other students.";
```
yousseftopaji, line 10: It is very good that you are experimenting with your code. I appreciate that :)
However, "return" would not log the message in the console. console log is a better choice for this level!

javascript/javascript1/week2/weather-wear.js
yousseftopaji, line 1: Very good job.

yousseftopaji (changes requested):
It is very good code all in all, keep up the good work!
Though, there are a few things that should be changed in the student-manager file.
I left a few comments about what must be changed.
Have a good weekend!

PR #88, opened 2024-11-21
ExercisesDone

Aymanngu:


javascript/javascript1/week2/FlightBookingFullnameFunction.js
```diff
@@ -0,0 +1,20 @@
+function getFullname(firstname, surname, useFormalName, gender) {
+    if (useFormalName === true) {
+        if (gender === "male") {
+            return "Lord " + firstname + " " + surname;
+        } else if (gender === "female") {
+            return "Lady " + firstname + " " + surname;
```
julczka, line 6: I'd just return `firstname + " " + surname;` if there is no gender of gender is something else then the two values we check for (people can type M or F for example)

julczka (approved):
Very nice! Good job!

PR #91, opened 2024-11-23
homework

Aminata00:


javascript/javascript1/week2/event.js
```diff
@@ -0,0 +1,18 @@
+function getEventWeekday(daysFromToday) {
+    const weekdays = ["Sunday", "Monday", "Tuesday", "Wednesday", "Thursday", "Friday", "Saturday"];
+    const today = new Date().getDay();
+
+    let eventDay = today;
+    for (let i = 0; i < daysFromToday; i++) {
```
NVedsted, line 6: Creative use of a loop! ➰ Can you think of a way to accomplish this with [the remainder operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Remainder) instead? 😄

javascript/javascript1/week2/smartease.js
```diff
@@ -0,0 +1,38 @@
+function getFullName(formalName, firstName, surname) {
+    let fullName = ""; 
+    if (formalName === true ) {
+        fullName = "Lady " + firstName + " " + surname;
+      } else if (formalName === true ) {
+        fullName = "Lord " + firstName + " " + surname;
+      } else {
+        fullName = firstName + " " + surname };
```
NVedsted, line 8: This is great work, but I am having a hard time reading it. 😢 It is important to pay attention to the formatting of your code; for your own and others' sake. In this case, I erroneously believed at first that you only 
`return` in the `else` branch since the closing brace was being tricky. 😄 

Your editor can very likely help you; give it a google if you're unsure how to do it. 💪

javascript/javascript1/week2/studman.js
```diff
@@ -0,0 +1,23 @@
+const class07Students = [];
+const maxStudents = 6;
```
NVedsted, line 2: Very nice use of constant to store this value! 🌟 As a brief thought experiment, can you explain why this is good and how it helps you and others maintain code? 😄

javascript/javascript1/week2/studman.js
```diff
@@ -0,0 +1,23 @@
...
+
+function addStudentToClass(studentName) {
+    if (!studentName) {
+        return "Student name cannot be empty.";
+    }
+
+    if (class07Students.includes(studentName)) {
+        return "Student" + studentName +  "is already in the class.";
```
NVedsted, line 10: There are some formatting issues that are a bit hard to catch with this syntax; I'd recommend going with [template strings](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals) instead of string concatenation in these instance. 😄 

The user might also expect that the messages are printed to console directly as they could very easily miss the error here if they do not log the return of the function. 🚨

javascript/javascript1/week2/studman.js
```diff
@@ -0,0 +1,23 @@
...
+    }
+
+    if (class07Students.length >= maxStudents && studentName === "Queen") {
```
NVedsted, line 13: Did you remember to test all the cases highlighted in the homework? 🧪 I think our Queens will be very disappointed if they try to sign up for a full class. 🥺 On the contrary, everyone else will be delighted!  Can you spot the bug? 🐛

javascript/javascript1/week2/weatherwear.js
```diff
@@ -0,0 +1,7 @@
+function getOutfit  (temp){
+    let outfit = "";
+    if (temp < 10 || temp === 10 ) {return outfit = "Better  get in your boots and jacket !"}
```
NVedsted, line 3: In these lines, you return an assignment to `outfit`. Can you explain why this works? 🤔 (It is a bit weird so it is fine if you cannot!)

In this case, you could return the string directly instead and avoid the variable `outfit` altogether. Can you spot other places in your homework for this week where you can remove a variable? 😄

NVedsted (changes requested):
Good job with the homework! ⭐ I have put a few comments on your submission mostly with alternative ideas to make life easier for you in the future. 😄 There are some bugs with your student manager though, so I would spend most of your follow-up time there! 🙂

Aminata00:
> Good job with the homework! ⭐ I have put a few comments on your submission mostly with alternative ideas to make life easier for you in the future. 😄 There are some bugs with your student manager though, so I would spend most of your follow-up time there! 🙂

// thank you so much for the clear review 😊! I'll make the changes ASAP and quote you again maybe ?

javascript/javascript1/week2/event.js
Aminata00: I have a hard time with remainders and that is why I went the long way haha :( But I can try👀

javascript/javascript1/week2/studman.js
Aminata00: I think that storing the maximum number of students into one variable makes it easier to find in the code (with the name😂) and make changes if necessary 🤔

javascript/javascript1/week2/studman.js
Aminata00: oh no let me fix it and save the queen😓

PR #93, opened 2024-11-25
Javascript_1, week_3 homework, Andrii Pavliuk

andrpavl:


javascript/javascript1/week3/ex-2.js
```diff
@@ -0,0 +1,16 @@
+//  Exercise 2: When will we be there??
+
+const travelInformation = {
+	speed: 50,
+	destinationDistance: 432,
+};
+
+function calculateTravelTime(travel) {
+	const totalTravelMinutes = (travel.destinationDistance / travel.speed) * 60;
```
NVedsted, line 9: Very nice work; especially the variable naming! 🤩 

To be a bit annoying, if something took a minute, it would return "Travel time is: 0 hours and 1 minutes". Can you think of ways to improve this? 😄

javascript/javascript1/week3/ex-1.js
```diff
@@ -0,0 +1,20 @@
...
+	"Rasmus",
+	"Samuel",
+	"katrine",
+	"Tala",
+];
+const nameToRemove = "Ahmad";
+
+// Write some code here
+
+const indexOfNameToRemove = names.indexOf(nameToRemove);
```
NVedsted, line 15: Excellent! ⭐

javascript/javascript1/week3/ex-4.js
```diff
@@ -0,0 +1,40 @@
...
+
+const notes = [];
+
+function saveNote(content, id) {
+	const note = { content, id };
+	notes.push(note);
+}
+
+function getNote(id) {
+	for (let i = 0; i < notes.length; i++) {
```
NVedsted, line 11: Nice and readable implementation! ⭐ The assignment also wants you to come up with a new feature which seems to be missing here. 🔍 

You have used `map` extensively until now; can you find a similar method that could solve this without a for loop? 😄

javascript/javascript1/week3/ex-5.js
```diff
@@ -0,0 +1,28 @@
...
+}
+
+function showStatus(arr = activities) {
+	let totalDuration = 0;
+	arr.map(({ duration }) => (totalDuration += duration));
+	console.log(
+		`You have added ${activities.length} activities. They amount to ${totalDuration} min. of usage`
+	);
+
+	totalDuration > usageLimit
```
NVedsted, line 18: Good job on making this as concise as it is! ⭐ The assignment also asks for a new feature to be implemented, but I will not ask that of you as the task is optional; unless you want to, of course! 😛 

A minor styling nitpick is that I would prefer an if statement here since the branches of your ternary are statements and not expressions; this is a personal opinion though. Alternatively, the ternary could also be moved into the `console.log` so it returns a string instead. 😄

javascript/javascript1/week3/ex-3.js
```diff
@@ -0,0 +1,68 @@
...
+		title: "Narcos",
+		days: 31,
+		hours: 12,
+		minutes: 55,
+	},
+];
+
+function logOutSeriesText(series = seriesDurations) {
+	let totalTime = 0;
+	const lifeInMinutes = 80 * 365 * 24 * 60;
```
NVedsted, line 56: Good use of a constant; makes it nice and readable! 👍

NVedsted (changes requested):
Excellent hand-in! 🌟 Generally, easy to review your code; kudos! 😄 

You seem to have grown fond of `map` which is nice! It is a bit early for such features, but if you understand higher-order functions and lambdas, then it is good practice to use them already now.

I will highlight that it seems you are using `map` to loop through each element, here `forEach` would be more appropriate. You are often using this to sum values together where something like [`reduce`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/reduce) is much more appropriate. I would you to replace your `map`s with `forEach` or, if you're up for it, `reduce` (when it makes sense!). 😄 

PR #94, opened 2024-11-26
Javascript javascript1 week3/parisa

SeyedehParisaMousaviamiri:


javascript/javascript1/week3/warmUp.js
```diff
@@ -0,0 +1,75 @@
...
+  ];
+function calculateSeriesTimePercentage(seriesDurations, averageLifespanYears) {
+    const averageLifespanMinutes = averageLifespanYears * 365 * 24 * 60; // Convert lifespan to minutes
+    let totalPercentage = 0;
+  
+    seriesDurations.forEach((series) => {
+      const seriesMinutes = series.days * 24 * 60 + series.hours * 60 + series.minutes; // Convert series time to minutes
+      const seriesPercentage = (seriesMinutes / averageLifespanMinutes) * 100;
+      totalPercentage += seriesPercentage; // Add to total
+      console.log(`${series.title} took ${seriesPercentage}% of my life`);
```
NVedsted, line 68: Phew, that's a lot of decimals for the percentages! 😰 Perhaps only showing a few decimals would suffice? 😄

javascript/javascript1/week3/note.js
```diff
@@ -0,0 +1,46 @@
...
+      console.log(`The note with id: ${note.id}, has the following note text: ${note.content}`);
+    });
+  }
+  
+  logOutNotesFormatted();
+
+  //Set Reminders for Notes
+
+  function setReminder(id, reminderTime) {
+    const note = notes.find((note) => note.id === id);
```
NVedsted, line 39: You've switched a bit between for loops and higher-order functions such as `forEach` and `find`. How did you pick which approach to use? 😄 

Can you briefly explain what `find` and `=>` does? 🔍

javascript/javascript1/week3/note.js
```diff
@@ -0,0 +1,46 @@
...
+}
+saveNote( "Pick up groceries", 1);
+saveNote( "Do laundry", 2);
+
+console.log(notes);
+
+//Get a note
+
+function getNote(id) {
+    const findNote = notes.find((note) => note.id === id);
```
NVedsted, line 19: Is this variable strictly needed? 😄

NVedsted (approved):
Nice hand-in! ⭐ I have left a few comments for you. As a general remark, I suggest going with `for...of` loops instead of `for (let i = ...)` as they are easier to type and less error-prone! 😄 

javascript/javascript1/week3/warmUp.js
SeyedehParisaMousaviamiri: Yes, it is a little complicated. I googled and just found this solution!

javascript/javascript1/week3/note.js
SeyedehParisaMousaviamiri: Do you mean const note? I think I need to declare const note as an object to show that the array notes are an array of objects.  Is not correct?

javascript/javascript1/week3/note.js
SeyedehParisaMousaviamiri: I pick find and forEach method in the list of the mdn website. I found out forEach is easier to write compare for :)
find is a method to recall the first element in which condition suits it. F.X,  notes.find(note) means find the note in the array notes. But we need a condition to find properly so we used "=>".

SeyedehParisaMousaviamiri:
Thanks for spending time to review my assignment :) I appreciate it. Many Thanks.

javascript/javascript1/week3/note.js
NVedsted: I am referring to `findNote` that is being immediately returned. 😄

SeyedehParisaMousaviamiri: Oh, yes. The findNote variable in my function searches through an array of notes to find a specific note that matches the given ID. Would you happen to have another suggestion for finding the particular note?

PR #95, opened 2024-11-26
add homework

najmeh25:


javascript/javascript1/week3/seriesLife.js
```diff
@@ -0,0 +1,53 @@
...
+    minutes: 0,
+  },
+];
+
+function logOutSeriesText() {
+  const totalLifeMinutes = 80 * 365 * 24 * 60;
+  console.log("Total Life Minutes: " + totalLifeMinutes); 
+  let totalPercentage = 0; 
+
+  for (const series of seriesDurations) {
```
NVedsted, line 27: Very nice use of `for...of`! You have used `for (let i = 0;...)` in other places, what made you choose differently here? 😄

Generally, I would suggest you stick to this form as often as possible. Not only is it shorter to type out, it is also less error prone! 🙂

javascript/javascript1/week3/seriesLife.js
```diff
@@ -0,0 +1,53 @@
...
+
+     if (isNaN(seriesMinutes)) {
+      console.log("Invalid data for series:", series);
+      continue; 
+    }
+
+    const percentage = (seriesMinutes / totalLifeMinutes) * 100;
+    console.log(`Percentage for ${series.title}: ${percentage}`); 
+    if (isNaN(percentage)) {
+      console.log("Error: NaN detected for percentage calculation.");
```
NVedsted, line 40: This error would probably confuse whoever called your function; can you think of a way to avoid this altogether? (Hint: why would `percentage` ever be `NaN`? And what could we do instead in that case?) 😄

javascript/javascript1/week3/showStatus.js
```diff
@@ -0,0 +1,36 @@
...
+}
+
+function showStatus() {
+  if (activities.length === 0) {
+    console.log("Add some activities before calling showStatus");
+    return;
+  }
+
+  let totalDuration = 0;
+  activities.forEach(activity => {
```
NVedsted, line 20: This hand-in has a mix of for-loops and more advanced features like `forEach`. Are there any particular reason for the switch back and forth? 😄

Also can you briefly explain what `forEach`and `=>` does? 🙂

Remember to only move onto functions like `forEach` once you are comfortable with for loops and higher-order functions! If you already are, then I'd recommend looking into reduce for this case as it is more fitting. 🧠

javascript/javascript1/week3/travel.js
```diff
@@ -0,0 +1,21 @@
...
+};
+
+function calculateTravelTime(travelInfo) {
+  const { speed, destinationDistance } = travelInfo;
+
+  if (speed <= 0) {
+    return "Error: Speed must be greater than 0";
+  }
+
+  const totalTimeInHours = destinationDistance / speed; // زمان به ساعت
```
NVedsted: I would suggest keeping your code, including comments, in English as it is the most accessible. 😄

NVedsted (changes requested):
This hand-in seems very decent and well-coded, but the structure of the files has made it very hard to review. 🥺 I have tried my best and added a few comments. In the future, ensure you name your files correctly, avoid splitting tasks into multiple files, and avoid duplicating code between files; your reviewer will thank you! 😄 

najmeh25:
Hi,Thank you very much for guiding me, I will fix all the problems and send
it again.

PR #97, opened 2024-11-26
Javascript javascript1 week3/farzaneh

FarzanehAhmadi:


javascript/javascript1/week3/warmup1.js
```diff
@@ -0,0 +1,22 @@
...
+  "kristina",
+  "Rasmus",
+  "Samuel",
+  "katrine",
+  "Tala",
+];
+const nameToRemove = "Ahmad";
+
+
+function removeName(arr,name){
```
NVedsted, line 14: Nice and concise; good job! ⭐ (Small nitpick: you are technically not using `nameToRemove`!)

javascript/javascript1/week3/warmup2.js
```diff
@@ -0,0 +1,14 @@
+const travelInformation = {
+  speed: 50,
+  destinationDistance: 432,
+};
+
+function travelDuration(info){
+  const time = info.destinationDistance / info.speed;
+  const hour = Math.floor (time);
+  const min = Math.trunc((time - hour)*60);
+  return `${hour} hours and ${min} minutes`
```
NVedsted: The variables are generally well-named here, but `time` could do with a unit and `min` can be misunderstood as `minimum`. 👀 

To be a little annoying, if something took exactly 1 minute, you would return `0 hours and 1 minutes`; can you think of a way to improve this? 😄

javascript/javascript1/week3/warmup3.js
```diff
@@ -0,0 +1,39 @@
...
+const myAge = 29;
+
+function calculateLifeSpentOnSeries(seriesArray, age){
+  const ageInHour = age * 365.25 * 24;
+  const result = [];
+
+  for (let i = 0; i<seriesArray.length; i++){
+    const seriesInHour = seriesArray[i].days * 24 + seriesArray[i].hours;
+    const percentage = (seriesInHour / ageInHour *100).toFixed(4);
+    result.push(`${seriesArray[i].title} took ${percentage}% of my life`)
```
NVedsted: Can you think of a way to simplify this and the printing logic below? 🔍 I assume your motivation for putting it into an array is to have line breaks; if so, can you think of another way to do so? 😄

javascript/javascript1/week3/notepad.js
```diff
@@ -0,0 +1,64 @@
...
+      notes[i].completed = true;
+    }
+  }
+  return notes;
+}
+markTaskCompleted(2)
+console.log(notes)
+
+//Checking if a task is completed or not
+function checkTaskCompleted(content){
```
NVedsted: Good job with your unique feature! ⭐ 

What would happen if a developer in the future set `note.completed` to `false`? 🐛 Would that behave as you expect?

In JavaScript, `undefined` is often a nuisance that wreak havoc on your code base. I recommend avoiding relying on it if you can get away with it. Perhaps `note.completed` should be set to a value when notes are added? 😄 

As a minor note, it might make more sense to check by ID instead of content, but I assume this was a deliberate choice? 👀

javascript/javascript1/week3/smartphone-usage-app.js
```diff
@@ -0,0 +1,66 @@
...
+  } else{
+    return `Add some activities before calling showStatus.`;
+  } 
+}
+
+//Test cases
+let myStatus = showStatus();
+console.log(myStatus);
+
+addActivity("Call of Duty", 90);
```
NVedsted, line 41: You add 90 minutes of activity here, but it seems your `console.log` says that the total amount of activities have remained unchanged and that the total have increased by 180 minutes and not 90. Can you see what goes wrong? 🐛 

```
You have added 2 activities. They amount to 90 min. of usage.
You have added 2 activities. They amount to 270 min. of usage.
```

(Hint 1: it is two separate issues.) (Hint 2: Avoid globals! 🌐)

javascript/javascript1/week3/smartphone-usage-app.js
```diff
@@ -0,0 +1,66 @@
...
+//Extra feature
+//Adding new key value pair to indicate activities category
+const appCategories = {
+  socialMedia: ['Instagram', 'WhatsApp', 'Facebook', 'Twitter', 'Snapchat', 'LinkedIn', 'TikTok', 'Reddit', 'Pinterest', 'WeChat'],
+  gaming: ['PUBG Mobile', 'Call of Duty', 'Candy Crush', 'Minecraft', 'Pokémon GO', 'Among Us', 'Clash Royale', 'Genshin Impact', 'Asphalt 9', 'Roblox'],
+  productivity: ['Google Calendar', 'Trello', 'Notion', 'Evernote', 'Microsoft To Do', 'Zoom', 'Slack', 'Google Docs', 'Forest', 'Habitica'],
+  entertainment: ['Netflix', 'Spotify', 'Disney+', 'YouTube', 'Twitch', 'Amazon Prime Video', 'Hulu', 'Audible', 'Apple Music', 'Crunchyroll']
+};
+
+function categorizeActivity(activities,categoies){
```
NVedsted, line 56: I like this a lot; awesome! 🤩

NVedsted (approved):
Very nice work! ⭐ I have added a few comments to your code. 😄 

Generally, I suggest you use `for...of` syntax as often as possible. The "old"/dumber for loop with `i` and math is error prone whereas `for...of` is simpler and harder to mess up. 🙂

javascript/javascript1/week3/warmup2.js
FarzanehAhmadi: Good point! 😊 I've updated the return statement so that if the hours are 0, it will return the minutes; otherwise, it will include both hours and minutes. I also renamed the variables for clarity, changing hours to totalHours and min to timeInMinutes.

javascript/javascript1/week3/warmup3.js
FarzanehAhmadi: Thanks for the feedback! 😊 I was considering using console.log inside the loop, but from what I understand, returning the result is a better approach. I'm not entirely sure, though. Could you help me understand when it's better to use return versus console.log?

javascript/javascript1/week3/notepad.js
FarzanehAhmadi: Thank you⭐ I see what you mean about undefined, so I've initialized note.completed as false when adding notes.

I used content to check tasks because I thought it would be easier to remember tasks by name (like checking if I washed the dishes) 😅, but I see that using ID might be more reliable.

FarzanehAhmadi:
Hi Nicklas,

Thanks for your help and feedback! They were very instructive, and I really appreciate it. I'll try to use more for...of loops instead of the old version. 😊
Best,
Farzaneh

javascript/javascript1/week3/warmup3.js
NVedsted: It is very much dependent on what the intention of the function is. 🙂  `console.log` is often seen as a "code smell" (aka "you're probably doing something weird" if you have it in your code), so you will almost always want to stick with returning stuff; this also gives the most flexibility to users of your function. E.g., in this context, the caller might want to show the text in the console or maybe put it into the HTML. So returning the message here is perfectly fine! 😄 

If you want to stick with returning here and avoid the array, then you could use a string which you concatenate to and use `\n` to make line breaks. Something like this:
```js
function calculateLifeSpentOnSeries(seriesArray, age) {
  let message = "";
  // ...
  for (let i = 0; i < seriesArray.length; i++) {
    // ...
    message += `${seriesArray[i].title} took ${percentage}% of my life\n`;
  }
  return message;
}
```

This is a bit of a contrived example and will also result in a loose `\n` in the end which is not super nice. Generally, such a function as this would not make any assumptions about the formatting but instead return structured data such as an object and then the caller can decide what to do. 😄

javascript/javascript1/week3/notepad.js
NVedsted: I can definitely agree with that, haha. In practice, you would probably have a search function you could get the ID with or perhaps have the notes listed in a grid where the user clicks (where the ID would also be available to us then without the user having to know it). 😄

PR #98, opened 2024-11-27
Javascript javascript1 week3/guzide

GuzideGuzelbey:
Warm-up exercises and the mandatory task are completed.
I have PD3(Danish language) exam on Monday, so if I can find time I will complete the optional task, as well.
Thanks for your help :)

### DIFF
--- a/javascript/javascript1/week3/week3-guzide/NoNOYes/NoYes.js
+++ b/javascript/javascript1/week3/week3-guzide/NoNOYes/NoYes.js
@@ -52,3 +52,20 @@ function logOutNotesFormatted() {
 }
 
 logOutNotesFormatted();
+
+//new feature: Remove Note
+
+function removeNote(content) {
+  for (let i = 0; i < notes.length; i++) {
+    if (notes[i].content === content) {
+      const deletedNote = notes.splice(i, 1)[0];
+      return console.log(
+        "The note with content: " + deletedNote.content + " is deleted."
+      );
+    }
+  }
+  console.log("Error: Note with content " + content + " not found.");
+}
+
+removeNote("Do laundry");
+removeNote("Find note");
--- a/javascript/javascript1/week3/week3-guzide/NoNOYes/NoYes.js
+++ b/javascript/javascript1/week3/week3-guzide/NoNOYes/NoYes.js
@@ -22,11 +22,11 @@ console.log("--------------------------");
 function getNote(id) {
   if (id === undefined || typeof id !== "number") {
     console.log("Error: Invalid or Missing id");
-    return null;
+    return;
   }
-  for (let i = 0; i < notes.length; i++) {
-    if (id === notes[i].id) {
-      return notes[i];
+  for (const note of notes) {
+    if (id === note.id) {
+      return note;
     }
   }
 }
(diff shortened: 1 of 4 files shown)


javascript/javascript1/week3/week3-guzide/warmups/script.js
```diff
@@ -0,0 +1,84 @@
...
+    title: "Dark",
+    days: 4,
+    hours: 16,
+    minutes: 40,
+  },
+];
+
+function logOutSeriesText(seriesDurations) {
+  // write code here
+  const lifeSpanMinutes = 80 * 365 * 24 * 60;
```
NVedsted, line 60: Good use of a constant! ⭐

javascript/javascript1/week3/week3-guzide/warmups/script.js
```diff
@@ -0,0 +1,84 @@
...
+console.log(names);
+console.log("-----------------------------------------");
+
+//When will we be there??
+const travelInformation = {
+  speed: 50,
+  destinationDistance: 432,
+};
+function travelDuration(travelinfo) {
+  const rawTravelDuration = travelinfo.destinationDistance / travelinfo.speed;
```
NVedsted: This is a bit verbose but gets the job done! ⭐ If you have the time, I'd suggest looking into getting this down to a fewer lines. 😄 Feel free to reach out if you get stuck!

javascript/javascript1/week3/week3-guzide/warmups/script.js
```diff
@@ -0,0 +1,84 @@
...
+  "kristina",
+  "Rasmus",
+  "Samuel",
+  "katrine",
+  "Tala",
+];
+const nameToRemove = "Ahmad";
+
+// Write some code here
+let removeElement = names.indexOf(nameToRemove);
```
NVedsted: Good job! 😄 As a small nitpick, good naming of variables will only get more important and here some might think the have the element/name itself but it is the index of the name. Can you think of a name that might be less confusing? 🙂

javascript/javascript1/week3/week3-guzide/NoNOYes/NoYes.js
```diff
@@ -0,0 +1,71 @@
...
+        notes[i].content
+    );
+  }
+}
+
+logOutNotesFormatted();
+
+//new feature: Remove Note
+
+function removeNote(content) {
```
NVedsted: Excellent note implementation and new feature! 🤩 

As a suggestion, I would find it more natural to delete by ID as it can be hard to get content right (e.g., casing, spacing, dots, etc. causing issues). This could be paired with a search function so the workflow would be: search for note, get the note, and delete it by ID. I do not expect you to implement this. 😄

NVedsted (changes requested):
Very nice hand-in! 🌟 I have added a few comments I would like you to look at. 😄 

Two general things:
 - I would like you to use template strings whenever you can! A general rule is whenever you find yourself `+`'ing strings together, then you should use a template string instead. 🙂
 - You seem to have a solid grasp of for loops; you might even have gotten annoyed by how much code you need to type just to go through an array. 🥴 I would like you to generally use [`for...of`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...of) whenever applicable as it is easier to use and less error-prone. 😄

javascript/javascript1/week3/week3-guzide/warmups/script.js
GuzideGuzelbey: thank you :)

javascript/javascript1/week3/week3-guzide/warmups/script.js
GuzideGuzelbey: More logical variable name is used. Thank you :)

GuzideGuzelbey:
First of all I need to apologise for the late response.
for...of loops and template strings integrated to overall code for simplicity.
Thanks for your constructive feedbacks and recommendations :)

PR #99, opened 2024-11-27
Javascript javascript1 week3/gayathri

GayathriVenkatraman:


javascript/javascript1/week3/note taking app/notetaking.js
```diff
@@ -0,0 +1,56 @@
...
+        note.content
+      } and it is ${note.completed ? "completed" : "not completed"}.`
+    );
+  }
+}
+
+logOutNotesFormatted();
+
+function markNoteAsCompleted(id) {
+  const note = notes.find((note) => note.id === id);
```
NVedsted, line 44: Great work on the notes implementation! ⭐ Until this point you have been using a mix of old-school for loops and `for...of`, but here you've used `.find` instead. Why did you make this choice? 😄 

As this is also a somewhat advanced feature, I want to make sure you fully understand it! Can you explain what `find` and `=>` does? 🙂

javascript/javascript1/week3/smart phone uasge app/smartphone.js
```diff
@@ -0,0 +1,61 @@
...
+}
+
+console.log(showStatus());
+
+addActivity("LinkedIn", 35);
+
+console.log(showStatus());
+console.log(activities);
+
+function getMostTimeSpentActivity() {
```
NVedsted, line 54: Excellent and cleanly formatted implementation and I like the feature you decided to add! 🤩 

If I were to run `addActivity("Youtube", 10)` after adding LinkedIn above. Would you still expect LinkedIn to be returned by this method? And why? If a change is needed, what would you do? 😄 (Mainly a thought exercise, so you do not have to implement unless you want to, of course.)

javascript/javascript1/week3/warmup_exercise/script.js
```diff
@@ -0,0 +1,83 @@
...
+  "Rasmus",
+  "Samuel",
+  "katrine",
+  "Tala",
+];
+const nameToRemove = "Ahmad";
+
+for (let i = 0; i < names.length; i++) {
+  if (names[i] === nameToRemove) {
+    names.splice(i, 1);
```
NVedsted: Very nice to consider that the name could be present multiple times! ⭐ As a small thought experiment, if you only were to remove one, can you come up with a good method on arrays that can help you get the index you need to splice? 😄

javascript/javascript1/week3/warmup_exercise/script.js
```diff
@@ -0,0 +1,83 @@
...
+];
+
+function logOutSeriesText() {
+  const averageLifespanInYears = 80;
+  const hoursInAYear = 365 * 24;
+  const totalLifespanInHours = averageLifespanInYears * hoursInAYear;
+
+  let totalPercentage = 0;
+
+  seriesDurations.forEach((series) => {
```
NVedsted, line 84: If you feel comfortable with these higher-order functions and lambdas, then the array method `reduce` would be the perfect one to use instead of `forEach` here! 😄

javascript/javascript1/week3/warmup_exercise/script.js
```diff
@@ -0,0 +1,83 @@
...
+const travelInformation = {
+  speed: 50,
+  destinationDistance: 432,
+};
+
+function calculateTime(speed, destinationDistance) {
+  let time = destinationDistance / speed;
+  const hours = Math.floor(time);
+  const minutes = Math.round((time % 1) * 60);
+  return `${hours} hours ${minutes} minutes`;
```
NVedsted: If the travel time was just one minute, you would return "0 hours and 1 minutes". Can you think of a way to improve this? 😄

NVedsted (approved):
Very nice and, in particular, neatly structured hand-in; awesome! ⭐ I have added some comments on your code that I'd like you to take a look at. 😄 

As a general note, I can see you generally have adopted `for...of` and higher-order functions such as `find` and `forEach`, which is great! Be sure you feel comfortable with writing "normal" loops before switching to the higher-order methods. 🧠 

There is a single-instance of an old-school for loop with `i` and math that I'd love for you to turn into one of the above as they are less error-prone. 🙂

javascript/javascript1/week3/note taking app/notetaking.js
GayathriVenkatraman: Thanks for this question. During our last class, we learned the differences between **find** and **filter**, and we also did some examples. So I wanted to try using **find** in the above function. 

find: It returns the first element that satisfies the given condition
=> : It is called as Arrow function, which is a simplest form of writing a traditional function. It is most commonly used in filter(), find() :)

Hope I have good understanding of this

javascript/javascript1/week3/warmup_exercise/script.js
GayathriVenkatraman: Okay, thanks for guiding me in this. I have used indexOf method to find the index of first occuring element and then splice, which will remove only one "Ahmad" from the array, not multiple "Ahmad".

javascript/javascript1/week3/warmup_exercise/script.js
GayathriVenkatraman: Good thought :) I have updated my code to display as 1 minute/ 1 hour(instead of hours).

javascript/javascript1/week3/smart phone uasge app/smartphone.js
GayathriVenkatraman: I have done few changes to the code which still shows LinkedIn as my mostTimeSpentActivity, after adding Youtube, 10.

javascript/javascript1/week3/warmup_exercise/script.js
GayathriVenkatraman: Sure I will try with using reduce method instead of forEach

GayathriVenkatraman:
Thanks a lot for your time on reviewing the code and also for the valuable feedback. It is really helpful and a way of learning & improving the code in different ways.

javascript/javascript1/week3/smart phone uasge app/smartphone.js
NVedsted: I don't think introducing another global variable here is a good idea as it just opens us up to getting the data out of sync; your old approach did a great job about this! 😄  What I was fishing after is that only a single activity is being considered, but if I logged the same activity multiple times perhaps the sum should be considered instead? E.g., if I logged 30 minutes on YouTube on Monday and then another 10 on Tuesday, then YouTube should probably win over LinkedIn as it has 40 total minutes compared to LinkedIn's 35 (in the old code).

I hope this makes any sense.

PR #100, opened 2024-11-27
javascript-javascript1-week3/jianxin

jianxinz233:


javascript/javascript1/week3/smartPhoneApp.js
```diff
@@ -0,0 +1,141 @@
...
+
+// as date format is already unified, then showStatus should be updated. Here I would like to show status of today or a specific day.
+// here I want to alert only for the usage limit of today, if the previous records sum over limit, it won't alert but a summary of usage.
+function showStatus(dateInput) {
+    if(activities.length === 0){
+        return 'Add some activities before calling showStatus.'
+    }else{
+        let activitiesOneDay = activities.filter(activities => activities.date === dateInput);
+        if(activitiesOneDay.length > 0){
+            let usageOfOneDay = activitiesOneDay.reduce((totalDuration, activities) => totalDuration + activities.duration, 0);
```
NVedsted, line 107: This is an interesting twist on the assignment! 😄 Can you briefly explain what `filter`, `reduce` and `=>` does? 🙂

It is a bit hard to follow the implementation though as there are many layers of if-statements. I would suggest looking into the "early-exiting" idiom; some quick googling should get you going! 😄 The idea is to reduce the nesting as much as possible by returning as soon as possible. A quick example is on line 104: since you returned in the then branch, you technically do not need to put the rest inside `else`. Can you see why? 🙂

javascript/javascript1/week3/smartPhoneApp.js
```diff
@@ -0,0 +1,141 @@
...
+                return `You have added ${activitiesOneDay.length} activities today. They amount to ${usageOfOneDay} min. of usage.`
+            }
+        }else{
+            return `Add some activities on ${dateInput} before calling showStatus.`
+        }
+    }
+}
+
+const checkToday = new Date().toLocaleDateString("en-US");
+console.log(showStatus(checkToday)); //alert when sum of today are over limit.
```
NVedsted, line 120: Excellent tests! 👍 I hope dates haven't scared you away; they can be terrible to work with. 🥴

javascript/javascript1/week3/warmUp.js
```diff
@@ -0,0 +1,79 @@
...
+
+
+//   When will we be there??
+const travelInformation = {
+    speed: 50,
+    destinationDistance: 432,
+};
+
+function calculateTravelTime(informationInput){
+    calculateTime = informationInput.destinationDistance / informationInput.speed
```
NVedsted, line 30: The use of `parseInt` and `Number` should not be necessary here. Can you see why? 🔍 

It is also important to remember to use `let` or `const` for your variables! Which would you choose here? 😄

javascript/javascript1/week3/smartPhoneApp.js
```diff
@@ -0,0 +1,141 @@
+// Adding an activity
```
NVedsted, line 1: It is nice to have the history and context for your decisions, but for the sake of the reviewer, please only submit your final result and avoid keeping commented-out code in. 😄

javascript/javascript1/week3/noteTakingApp.js
```diff
@@ -0,0 +1,65 @@
...
+function getLatestId() {
+  let latestId = notes[notes.length - 1].id;
+  console.log(`The latest ID is ${latestId}.`);
+}
+
+getLatestId();
+
+
+// Unique feature: What if user forgets the id, and just remember a keyword of content, how to find the note.
+function getNoteWithoutId(keyword) {
```
NVedsted, line 54: This is a very neat searching feature; good job! ⭐

NVedsted (changes requested):
Excellent hand-in! ⭐ I have left a few comments for you. 😄 In general, I would love to see `for...of` being used more extensively over their counterpart with indexes. It is easier to type out and less error-prone. 🙂

javascript/javascript1/week3/smartPhoneApp.js
jianxinz233: Yes!!! Almost two weeks, I was struggling with dates! This idea actually came from the last online session, when Thomas explained to me. Thank you both!

javascript/javascript1/week3/smartPhoneApp.js
jianxinz233: Sorry for that 🥺

javascript/javascript1/week3/smartPhoneApp.js
jianxinz233: Yes, `filter` and `=>` were once mentioned in the class, when we want to find an element not knowing if there are several or just one, using 'includes' may just find one element. But here we can have several activities in one day, so  I use `filter` here.
Then the `reduce()` is used to calculate the sum of duration. When I google how to calculate the sum of numbers in object and array, most of the answers are using `reduce((accumulator, currentValue) => accumulator + currentValue, 0)`, so I tried to use here.
As you mentioned, I also feel more and more difficult to have so many conditions together. I will go to spend some time on "early-exiting". 
Because I just had a project presentation yesterday, I just start homework today 😂

javascript/javascript1/week3/warmUp.js
jianxinz233: So the here is it better to use `Math.floor(calculateTime)` ?
Maybe `let` in the function?

NVedsted: Everything should already be and stay a number so `parseInt` and `Number` at best should do nothing or they might break stuff. I think putting a number into `parseInt` might convert it to a string and back to a number, but maybe it is smart! 😄

For `const`/`let`, I'd suggest always starting with `const` and then switch to `let` if you find yourself needing to modify it. (It is good practice to see if you can keep something `const` as it is easier to maintain code if things cannot change in weird places.) 👍

PR #101, opened 2024-11-27
Final

StefanyRomeroV:


javascript/javascript1/week3/script.js
```diff
@@ -0,0 +1,147 @@
...
+    "Samuel",
+    "katrine",
+    "Tala",
+];
+
+const nameToRemove = "Ahmad";
+
+function removeName(names, nameToRemove) {
+    const remove = names.indexOf(nameToRemove);
+    if (remove !== -1) {
```
NVedsted, line 19: Excellent implementation; I especially like the handling if the name does not exists. Good job! ⭐

javascript/javascript1/week3/script.js
```diff
@@ -0,0 +1,147 @@
...
+        hours: 12,
+        minutes: 0,
+    },
+];
+
+
+function logOutSeriesText() {
+    const averageLifespan = ((80 * 365 * 24 * 60) / 100); // assuming 80 years
+    let totalPercentage = 0;
+    seriesDurations.forEach((series) => {
```
NVedsted, line 76: This hand-in has a mix of for-loops and more advanced features like `forEach`. Are there any particular reason for the switch back and forth? 😄 

Also can you briefly explain what `forEach` and `=>` does? 🙂

Remember to only move onto functions like `forEach` once you are comfortable with for loops and higher-order functions! If you already are, then I'd recommend looking into `reduce` for this case as it is more fitting. 🧠

javascript/javascript1/week3/script.js
```diff
@@ -0,0 +1,147 @@
...
+
+saveNote("Pick up groceries", 1);
+saveNote("Do laundry", 2);
+
+console.log(notes);
+
+function getNote(id) {
+
+    if (id === undefined || typeof id !== "number") {
+        return "Error, Id have to be defined and have to be a number";
```
NVedsted, line 109: Really loving the error-handling; good job! 🤩

javascript/javascript1/week3/script.js
```diff
@@ -0,0 +1,147 @@
...
+        console.log(" The note with id: " + note.id + ", has the following note text: " + note.content)
+    }
+}
+logOutNotesFormatted();
+
+//Unique feature 
+//I would like to add remove note 
+
+function delateNote(id) {
+    const noteIndex = notes.findIndex(note => note.id === id);
```
NVedsted, line 135: You found notes by ID a bit differently in you `getNote` method; how come? 🙂

NVedsted (approved):
Very nice hand-in! ⭐ I have left a few comments for you. 🙂 Some general notes:
 - You've used `forEach` and `findIndex` instead of for loops in a few places, this is great if you're comfortable and fully understand how they work! 😄 
 - For the for loops, I would recommend using [`for...of`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...of) instead when applicable as it is less error prone and easier to type out! 🚀 
 - Lastly, I would also suggest using [template strings](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals) when you need to build strings. Basically, anytime you find yourself adding strings together, you should probably use a template string instead. 🙂

javascript/javascript1/week3/script.js
StefanyRomeroV: Hi, the mix is intentional because I'm trying to implement and solve the exercises in different ways to challenge myself but also to understand truly the reason for the code if that makes sense.

From what I understand, forEach is similar to a for loop but is better suited when I want to process or calculate something for each element in the same array. The => syntax represents a function that processes each element in the array.

For example, in seriesDurations.forEach((series) => { ... }), the code processes each series object in the seriesDurations array, calculating the time, percentage, or any other operation for every series.

Of course, this is how I understood it so far please correct me if I'm mistaken.

NVedsted: Thanks for the explanation! I would say you are pretty spot on. Although `forEach` is not super suited for processing data. In these cases you probably want to use map, filter, reduce, etc. It is mostly for "side effects" such as logging or sending data, i.e. affecting things outside the return value. We are getting a bit ahead of ourselves here so it is fine if this does not make sense. 😊

PR #102, opened 2024-11-27
Javascript-javascript1-week3/yuusuf

Ganja0003:


javascript/javascript1/week3/Note-taking-app.js
```diff
@@ -0,0 +1,40 @@
+const notes = [];
```
NVedsted, line 1: Nice implementation of note handling! 😄 Be careful with the formatting though. There is a lot of additional space in this file that make it a bit harder to read. 🥺

javascript/javascript1/week3/javascript-warmup/When-will-we-be-there.js
```diff
@@ -0,0 +1,18 @@
...
+  function calcTravelTime(travelInformation){
+
+    const hoursWithDecimals = travelInformation.destinationDistance / travelInformation.speed
+    const hoursWithOutDecimals = Math.floor(hoursWithDecimals);
+    const minutesWithDecimals = (hoursWithDecimals - hoursWithOutDecimals)* 60 ;
+    const minutesWithOutDecimals = Math.round(minutesWithDecimals);
+    console.log(`${hoursWithOutDecimals} hours and ${minutesWithOutDecimals} minutes`);
+  }
+  
+  const travelTime = calcTravelTime(travelInformation);
```
NVedsted, line 15: The variable declaration/assignment doesn't seem to do anything here and could be removed. 🙂

javascript/javascript1/week3/javascript-warmup/Series-duration-of-my-life.js
```diff
@@ -0,0 +1,57 @@
...
+  ];
+  
+  function logOutSeriesText() {
+    // write code here
+    const eightyYearsIntoHours = 80 * 365 * 24;
+
+
+
+    // gameOfThrones Variables
+    const gameOfThronesdaysIntoHours = seriesDurations[0].days*24;
```
NVedsted, line 29: This works as intended, but it has a lot of duplicated code, and what happens if a new series is added? 🔍 

I would like you to resolve these issues by implementing it with a loop instead. 😄

NVedsted (changes requested):
Nice hand-in! ⭐ I have left a few comments I would like you to take a look at. 😄 As a general note, I would suggest using [`for...of`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...of) when possible as it is easier to write and less error-prone. 🙂

PR #103, opened 2024-11-27
javascript-javascript1-week3/hossein

hossein1065:


javascript/javascript1/week3/homework.js
```diff
@@ -0,0 +1,169 @@
...
+addDte(1)
+addDte(2)
+
+// CactusIO-interactive (Smart phone usage app) optional
+  
+// Adding an activity
+
+let activities=[];
+let usagelimit=90;
+function addActivity(date=``,activity=``,duration){
```
NVedsted: What is the reason for adding default value for the two first arguments here? 😄 It might just create confusion as you will probably want to fill out `duration` no matter.

As a small nitpick, `usagelimit` could also be a constant here. 🙂

javascript/javascript1/week3/homework.js
```diff
@@ -0,0 +1,169 @@
...
+logOutSeriesText();
+
+
+//NOnoN0nOYes (Note taking app)
+// Save a note
+
+ const notes=[];
+ function saveNote(content, id){
+  notes.push({ content, id })
+  console.log(notes)
```
NVedsted: Excellent to see you using `console.log` to see the state as your program is running; a nice and easy debugging technique. 😄 I recommend going through your debug logs and clean them up before submitting code for review as it is likely not wanted in the final product. 🙂

javascript/javascript1/week3/homework.js
```diff
@@ -0,0 +1,169 @@
...
+Remove(names,`Ahmad`)
+console.log(names)
+
+
+//When will we be there??
+
+
+function travelTime(distance,speed){
+  return{
+    calculate(){
```
NVedsted: This looks more complex than it needs to. 😮 What is the reason for returning an object with a function instead of just the result right away? 🙂

javascript/javascript1/week3/homework.js
```diff
@@ -0,0 +1,169 @@
...
+  "kristina",
+  "Rasmus",
+  "Samuel",
+  "katrine",
+  "Tala",
+];
+console.log(names)
+const nameToRemove = "Ahmad";
+
+function Remove(arrayOfNames=[],name){
```
NVedsted: Be careful with naming! Many will assume that variables/functions/etc. starting with a capital letter is a class; this can also be seen on GitHub where the syntax color changes for this function only. It has no functional implication and is purely styling, but I would recommend sticking to lower-case initial letter for things that are not classes (covered later). 😄

javascript/javascript1/week3/homework.js
```diff
@@ -0,0 +1,169 @@
...
+ }
+ saveNote(`Pick up groceries`, 1);
+saveNote(`Do laundry`, 2);
+console.log(notes)
+
+
+// get a note
+function getnote(id){
+  for(i=0; i<notes.length;i++){
+    if (notes[i].id===id)
```
NVedsted: This if-statement has no braces while others do. I personally prefer sticking with always using braces as it is more consistent and you avoid gotchas. This is purely up to you though! What do you think? 😄

NVedsted (changes requested):
Super nice hand-in! ⭐ I have added a few comments I'd like you to take a look at. As for some general remarks:
 - Your code formatting is a bit hard to follow; I'd recommend running your code through a formatter before submitting. 😄 I must admit that I actually ran your code through my editor's formatter to be sure I read it correctly. 🥺 
 - Your hand-in contains a few for-loops that could instead be [`for...of` loops](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...of). I highly recommend those as they are less error-prone and you ultimately just get to type a lot less! 🚀 

javascript/javascript1/week3/homework.js
hossein1065: your right. I remove the default for the two arguments and added console,log() to make function output more readable and changed the let to const

javascript/javascript1/week3/homework.js
hossein1065: may you please give me more details and I write my code in VS code and run each part that is completed to see the result before submitting

javascript/javascript1/week3/homework.js
hossein1065: I just wanted to create a factory function to use this function for each distance and speed

javascript/javascript1/week3/homework.js
hossein1065: yes you right , I did not use the curly braces because I had only one statement on the if but it was better to used it

hossein1065:
I Thanks for your advice, I did what you recommended. I always run the code before posting it to see the result.

PR #105, opened 2024-11-28
Javascript-javascript1-week3/andriiK

pahanan:


javascript/javascript1/week3/Javascript_warmup.js
```diff
@@ -0,0 +1,110 @@
...
+  // Code done
+  
+  console.log(names); // ['Peter', 'Yana', 'kristina', 'Rasmus', 'Samuel', 'katrine', 'Tala']
+
+
+
+  // #2 Speed function 
+
+  function calculateTime({speed, destinationDistance}){
+    x = typeof(speed) === "number" && typeof(destinationDistance) === "number";
```
NVedsted, line 46: If you have complex boolean logic, it can be very good to break it up so it becomes more digestible and easy to read! 😄 It is paramount to pick good names for variables though! Can you find better names for `x` and `y` here? There is also an instance of this in the CactusIO implementation that I'd like you to revisit as well. 🙂

You are also missing `const` or `let`; which one would you choose here?

javascript/javascript1/week3/notes.js
```diff
@@ -0,0 +1,58 @@
...
+}
+
+
+
+saveNote("Pick up groceries", 1);
+saveNote("Do laundry", 2);
+
+console.log(notes); // [{content: 'Pick up groceries', id: 1}, {content: 'Do laundry', id: 2}]
+
+const firstNote = getNote(1);
```
NVedsted, line 37: If I try to `getNote(2)`, it doesn't work! 🐛 Can you see what goes wrong and fix it? 🔍

javascript/javascript1/week3/notes.js
```diff
@@ -0,0 +1,58 @@
...
+
+
+  
+logOutNotesFormatted(); // should log out the text below
+  
+// The note with id: 1, has the following note text: Pick up groceries
+// The note with id: 2, has the following note text: Do laundry
+
+function findByWord(word){
+    const regex = new RegExp(`\\b${word}\\b`, "i");
```
NVedsted, line 48: Now this is an interesting feature! 😀 Exciting to dip into Regex as well. It is a bit fragile though, e.g., if the developer puts any Regex syntax into the line it will break. (Try `findByWord(")")` and see what happens.)

To simplify things, perhaps consider looking for any notes that contain the input. Can you think of a simpler way of doing this while still making it case-insensitive? 😄 

If no notes is found, `0` is returned. This is a bit different from other methods and should be streamlined. 📏 

You are using a complex feature to find the notes matching your Regex, can you explain what `.find` does, and what `=>` means? 🔍

javascript/javascript1/week3/Javascript_warmup.js
```diff
@@ -0,0 +1,110 @@
...
+    "Peter",
+    "Ahmad",
+    "Yana",
+    "kristina",
+    "Rasmus",
+    "Samuel",
+    "katrine",
+    "Tala",
+  ];
+  const nameToRemove = "Ahmaad";
```
NVedsted, line 13: There is a typo here which breaks the example. 😢 Good job on the implementation though! 👍

javascript/javascript1/week3/CactusIO-interactive.js
```diff
@@ -0,0 +1,82 @@
+let activities =[];
+let limit = 180;
+let value = 0;
+var options = { weekday: 'long', year: 'numeric', month: 'long', day: 'numeric' };
+var today  = new Date();
```
NVedsted, line 5: `options` and `today` are unused and defined with `var`. Remember to clean-up before submitting! 😄

javascript/javascript1/week3/CactusIO-interactive.js
```diff
@@ -0,0 +1,82 @@
...
+    } else {
+        return "0 activities. Add few more."
+    }
+}
+
+addActivity('2023, 7 , 19', "Youtube", 30);
+showStatus(activities);
+addActivity('2023, 7 , 18', "Twitch", 60);
+showStatus(activities);
+addActivity('2023, 7 , 18', "CS2", 180);
```
NVedsted, line 74: If I split this 180 minutes activity into two activities of 90 minutes each, I do not get an error stating I have exceeded my daily limit. 🐛 Can you see what goes wrong? (Hint: something sneaky is going on in line 30!)

Having the variable `value`, which I assume is supposed to contain the sum of activities, makes the code a bit fragile. First and foremost, we are now storing duplicate data, i.e., `value` is computed based on what is in `activities`. This means they can get out-of-sync! 😱 I would avoid this variable and simply compute the sum when it is needed; it is fine four our purposes here! (Maybe introduce a function that does this so it can be reused!)

Generally, I would recommend limiting global variables, i.e., variables not inside functions, to a minimum as it can be hard to maintain code that has too many globals. Generally, you should strive to have functions only be determined by their inputs; if you want to read more about this try googling "side effects". 😄

NVedsted (changes requested):
Nice job with the homework! ⭐ There are a few issues that I'd like you to address. I have added a few more comments than I'd like, and some of them are also very long, so I do not expect you to get through all of them; but you're awesome if you do! 😄 

Some general notes:
 - The code formatting is not super consistent; I'd suggest getting your editor to format the code for you! (Google should be helpful if you are unsure how! 🔍)
 - I'd suggest leaning more into [early-exiting](https://medium.com/swlh/return-early-pattern-3d18a41bba8) to reduce indentation and making your code easier to follow. 🤓 
 - Excellent use of template strings! Keep that up! 🤩 
 - There are a few instances where you have used `==` or `!=` which is discouraged. (Do you recall why? 🧠)
 - `for` loops with math and `i` and all that is a bit of a mouthful and error-prone. Luckily, we got [`for...of` syntax](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...of) that can eliminate most of it! I'd like you to use that when possible. 😄

Feel free to reach out if you have any questions or issues. 😄

PR #107, opened 2024-11-28
javascript-javascript1-week3/juan

PAINj9:


javascript/javascript1/week3/homework/script.js
```diff
@@ -0,0 +1,28 @@
...
+  ];
+  const nameToRemove = "Ahmad";
+  const borrar = names.indexOf(nameToRemove);
+
+  
+  // Write some code here
+  
+  if (names.includes(nameToRemove)) {
+
+     names.splice(borrar, borrar);
```
yousseftopaji, line 19: Splice(param1, param2) removes elements from an array and, if necessary, inserts new elements in their place, returning the deleted elements.
@param1 — The zero-based location in the array from which to start removing elements.
@param2 — The number of elements to remove.

so if the name is `"Yana"`for example, the first and the second parameter in your code will be 2, and therefore you will get `"Yana"` and `"kristina"` removed

does is make sense?

javascript/javascript1/week3/homework/script.js
```diff
@@ -0,0 +1,28 @@
...
+
+    }
+
+
```
yousseftopaji, line 23: How ever if you want it to be fancy and return the same result, instead of `if statement`:

```suggestion
function nameToRemove(name) {
  names.splice(names.indexOf(name), 1);
}
```

And you can also try this
it might be fun ;)
```suggestion
function nameToRemove(name) {
  names.splice(names.indexOf(name), 1, "Simon");
}
```

yousseftopaji (changes requested):
Great job!
this is a very good implementation, however there is only one mistake, that needs to be fixed.
All the best!

PR #108, opened 2024-11-28
javascript-javascript1-week3/ayman

Aymanngu:


javascript/javascript1/week3/itemArrayRemoval.js
```diff
@@ -0,0 +1,18 @@
...
+    "kristina",
+    "Rasmus",
+    "Samuel",
+    "katrine",
+    "Tala",
+  ];
+  const nameToRemove = "Ahmad";
+  
+
+  const updatedNames = names.filter((name) => name !== nameToRemove);
```
NVedsted, line 14: This is a nice approach that we will see a lot more of in the future! 😄 However, the task asks you to remove the name from `names`, so creating a new array won't suffice.

javascript/javascript1/week3/noNoNoNoYes.js
```diff
@@ -0,0 +1,35 @@
+const notes = [];
+
+function saveNote(content, id) {
+  notes.push({ content, id });
+}
+
+function getNote(id) {
+  if (typeof id !== "number") {
+    console.log("Error: ID must be a number.");
```
NVedsted, line 9: Nice and concise implementation! ⭐  Very nice to sanity check the input here! Perhaps the same should be done when saving notes? 😄 

The task also asks you to introduce a new feature which seems to be missing. 🥺

javascript/javascript1/week3/seriesDurationOfMyLife.js
```diff
@@ -0,0 +1,23 @@
+const seriesDurations = [
+    { title: "Game of thrones", days: 5, hours: 14, minutes: 0 },
+    { title: "Prison Break", days: 2, hours: 18, minutes: 0 },
+    { title: "Gomorrah", days: 1, hours: 23, minutes: 0 },
+  ];
+  
+  const averageLifeSpanInMinutes = 80 * 365 * 24 * 60;
```
NVedsted, line 7: Another nice and readable implementation; awesome! 🤩 There is some issue with the formatting here where all lines except the first seem to be indented weirdly.

javascript/javascript1/week3/whenWillWeBeThere.js
```diff
@@ -0,0 +1,11 @@
+const travelInformation = {
+    speed: 50,
+    destinationDistance: 432,
+};
+
+const timeInHours = travelInformation.destinationDistance / travelInformation.speed;
```
NVedsted, line 6: Excellent variable naming! ⭐ The tasks did technically ask for a function to be implemented; I suspect that will be no issue for you to fix. 😄

NVedsted (changes requested):
Very nice hand-in! ⭐ There are a few things I have comments on. 😄 

Generally, I would recommend avoiding the `for` syntax where you set up your own `i` or similar in favor of [`for...of`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...of) when possible.

There are also higher-order functions available, like `filter` which you used for one of the tasks. If you feel comfortable with higher-order functions and lambdas, then I'd suggest using them instead of loops altogether. 😄

PR #109, opened 2024-11-28
homework

Aminata00:


javascript/javascript1/week3/cactus.js
```diff
@@ -0,0 +1,47 @@
...
+    const date = new Date();
+    const today = date.toISOString().split('T')[0]; // Format date as "YYYY-MM-DD"
+    const todayActivities = activities.filter(activity => activity.date === today);
+
+    if (todayActivities.length === 0) {
+        console.log("No activities logged today.");
+    } else {
+        console.log(`You have logged ${todayActivities.length} activities today.`);
+        console.log("Activities:");
+        todayActivities.forEach(act =>
```
NVedsted, line 23: Nice and easy-to-follow implementation! ⭐  This line stands out a bit, why did you choose to go with a `.forEach` here? 😄

javascript/javascript1/week3/note.js
```diff
@@ -0,0 +1,73 @@
...
+  notes.push({ content, id }); 
+  return `Note with id ${id} saved successfully!`;
+}
+
+console.log(saveNote('going on a run', 1)); 
+console.log(saveNote('meeting with friends', 2)); 
+
+// Get a note by id
+function getNote(id) {
+  for (let i = 0; i < notes.length; i++) {
```
NVedsted, line 13: Good work with this one as well! 😄 This file contains a lot of for loops where numbers are being fiddled with. I would recommend using `for...of` syntax especially when you do not actually need the index of the element. 🙂

javascript/javascript1/week3/removenames.js
```diff
@@ -0,0 +1,22 @@
...
+  const nameToRemove = "Ahmad";
+  
+  function removeName(names, nameToRemove) {
+    const index = names.indexOf(nameToRemove); 
+    if (index !== -1) { 
+      names.splice(index, 1); 
+    }
+  }
+  
+e
```
NVedsted, line 20: Doesn't get more concise than this; good job! ⭐ There is an issue on this line though; I assume this `e` slipped past! 😄

javascript/javascript1/week3/series.js
```diff
@@ -0,0 +1,50 @@
...
+      days: 6,
+      hours: 8,
+      minutes: 12,
+    },
+  ];
+  
+  const lifespan = 80; 
+  const lifespanMinutes = lifespan * 365 * 24 * 60; 
+  
+  function getPercentage() {
```
NVedsted, line 25: As you start working on more complex projects and with other people, it will quickly become apparent that naming is of immense importance!  Here I think people would expect to get a number back when it is called `getPercentage`. Can you come up with a name that better describes what the function does? 😄

javascript/javascript1/week3/whenwillwe-1.js
```diff
@@ -0,0 +1,15 @@
...
+    destinationDistance: 432, 
+};
+
+function getArrival() {
+    const totalTimeInHours = travelInformation.destinationDistance / travelInformation.speed;
+    const hours = Math.floor(totalTimeInHours);
+    const minutes = Math.floor((totalTimeInHours - hours) * 60);
+    const seconds = Math.round(((totalTimeInHours - hours) * 60 - minutes) * 60);
+
+    return `${hours} hours, ${minutes} minutes, and ${seconds} seconds`;
```
NVedsted, line 12: This is an excellent solution, but I do want to be a bit annoying and note that if it takes less than an hour, then I would not want to see `0 hours`. Furthermore, when there is only 1 of a time unit, then grammar is wrong (e.g., `1 hours`). 😄

How would you handle this? 🙂

NVedsted (approved):
Nice work! 🌟 Everything seems to be in order (except for a tiny, annoying `e` 😛); I have added some comments with some suggestions and questions.

Feel free to reach out if you have any questions or issues. 😄

PR #110, opened 2024-12-03
Javascript javascript1 week4/jalil

jalilhu:
took me 7 hours, 
lots of google
lots of trial and error
different methods

it is not perfect.

thanks for reviewing my code


javascript/javascript1/week4/voiceAssistant.js
```diff
@@ -0,0 +1,171 @@
+let clientName = "";
+const toDoList = [];
+function getReply(command) {
+  const assitantsave = {
+    addName(command) {
```
jmf-umbraco: Great use of additional functions!

As a small optimization to make the code slightly more readable at-a-glance, you could consider moving them out to live alongside the `getReply()` function. Currently, it's a little hard to tell immediately what that function does vs. what it delegates to other functions, since we have to scroll past these ones first.

One convention you'll often see is to put the main/important/public functions higher up in a file or class. Then any less important helper functions go lower down.

javascript/javascript1/week4/voiceAssistant.js
```diff
@@ -0,0 +1,171 @@
...
+      const name = command.split(" ").slice(-1)[0];
+      console.log("Nice to meet you " + name);
+      clientName = name;
```
jmf-umbraco: Consider assigning directly to `clientName` to remove the intermediate `name` variable.

javascript/javascript1/week4/voiceAssistant.js
```diff
@@ -0,0 +1,171 @@
...
+    },
+    viewMyName(command) {
+      if (clientName === "" || clientName === " ") {
```
jmf-umbraco: Is it actually possible for `clientName === " "` to happen? The only time we assign to the clientName variable it's using `const name = command.split(" ").slice(-1)[0];`, and the `.split(" ")` part should mean we don't end up with space characters in the assigned value.

That said, it's good practice to code "defensively", which means trying to anticipate errors or other ways in which the code might misbehave and coding in a way to minimize the impact of those.

Ultimately if you're 100% certain that some condition should never happen, then you can choose to omit that code and perhaps end up with something slightly simpler. If you're not certain, you can leave it in - it's a bit of a balancing act and a judgment call though. Sometimes you won't think something can happen, but it does!

jmf-umbraco: Also, if it's possible to end up with a single space in the name, is it possible to end up with more? If you want to be even more "defensive", then you could consider trimming whitespace from the string and then checking if there's anything left over.

javascript/javascript1/week4/voiceAssistant.js
```diff
@@ -0,0 +1,171 @@
...
+        console.log("you have not told me your name yet.");
+      }
+      console.log("Your name is " + clientName);
+    },
+    addToMyToDo(command) {
+      const match = command.match(/add (.*?) to my todo/);
```
jmf-umbraco: Great use of regex capture groups!

javascript/javascript1/week4/voiceAssistant.js
```diff
@@ -0,0 +1,171 @@
...
+      toDoList.push(match[1]);
+      console.log(match[1] + " added to your list");
+    },
+    removeFromMyToDo(command) {
+      const match = command.match(/remove (.*?) from my todo/);
+      let index = toDoList.indexOf(match[1]);
+      toDoList.splice(index, 1);
```
jmf-umbraco: Uh oh... We might have a problem!

What happens if the item that the user tries to remove wasn't found in the list? E.g. if the user runs this sequence of commands:

```
getReply("add going out for dinner to my todo");
getReply("add ordering dessert to my todo");
getReply("remove fishing from my todo");
getReply("what is on my todo");
```

javascript/javascript1/week4/voiceAssistant.js
```diff
@@ -0,0 +1,171 @@
...
+      const date = new Date();
+
+      const day = date.getDate();
+      const month = date.toLocaleString("default", { month: "long" });
+      const year = date.getFullYear();
+
+      console.log(`${day}. of ${month} ${year}`);
+    },
+    simpleCalculation(command) {
+      const match = command.match(/what is (\d+)\s*([\+\-\*\/])\s*(\d+)/);
```
jmf-umbraco: One other possibility is that we can extract the calculation from the string and run it using a function called `eval(...)`.

However, I like your method a lot better!

It's a bit more restrictive than the `eval()` approach, because it can only run pre-defined operations (i.e. from your switch statement below). `eval()` can run _any_ arbitrary code that the user enters.

But that's potentially a big security problem. If someone finds a way to trick users into entering something they didn't intend to, then `eval()` will just run it regardless. Sometimes this might not even need the user to be tricked into entering something - the input to `eval()` could come from anywhere, depending on how a site is coded.

So great work here, and definitely think twice before resorting to `eval()`!

javascript/javascript1/week4/voiceAssistant.js
```diff
@@ -0,0 +1,171 @@
...
+            result = num1 - num2;
+            break;
+          case "*":
+            result = num1 * num2;
+            break;
+          case "/":
+            result = num1 / num2;
+            break;
+          default:
+            result = "Invalid operator";
```
jmf-umbraco: Good stuff having a default condition in case something's gone wrong. If you want to enhance it further, you could consider making the message you show back to the user a bit more detailed, e.g. by suggesting how they could fix the problem.

Perhaps something like:
```
Invalid operator - Try again with a +, -, *, or / equation
```

jmf-umbraco:
Great work overall. Most of the suggestions I have are just that - they're things you _could_ do, rather than things you _should_. There is one potential bug in there that it would be great for you to look into, though!

jmf-umbraco (changes requested):

javascript/javascript1/week4/voiceAssistant.js
jalilhu: Thank you for the suggestion! I'll reorganize the functions.

javascript/javascript1/week4/voiceAssistant.js
jalilhu: Great point!

javascript/javascript1/week4/voiceAssistant.js
jalilhu: Thanks 😄

javascript/javascript1/week4/voiceAssistant.js
jalilhu: now that i think about it, it might never happen. but it is better to think about rare cases. so trim was a good idea.
thanks for highlighting it.

javascript/javascript1/week4/voiceAssistant.js
jalilhu: Thanks for catching that! I'll add a check to handle cases where the item isn’t in the list and provide feedback to the user.

javascript/javascript1/week4/voiceAssistant.js
jalilhu: never thought about it this way. 😄 thanks for explaining in details

PR #114, opened 2024-12-05
javascript-javascript1-week4/hossein

hossein1065:


javascript/javascript1/week4/homework.js
```diff
@@ -0,0 +1,123 @@
+
+const users=[];
+let fName=``;
+const listOfToDo=[];
```
Vladimir-malina, line 4: For small projects like this, it's not an issue.
But for any bigger ones try not to declare global variables if it is not necessary - in order not to polute a global scope
"users", "fName" you use only in the function below - so why not to move it there

javascript/javascript1/week4/homework.js
```diff
@@ -0,0 +1,123 @@
...
+function getReply(command){
+  // part one(getting and checking the name and give the appropriate response)
+  
+  
+  const constantExpression =`Hello my name is`
+  if(command.startsWith(constantExpression)){
+   const strArray=command.split(` `);
```
Vladimir-malina, line 11: command.split(` `) - met in few paces in this function - you can extract it into one variable and use it in every "if" conditional

javascript/javascript1/week4/homework.js
```diff
@@ -0,0 +1,123 @@
+
+const users=[];
+let fName=``;
+const listOfToDo=[];
+function getReply(command){
+  // part one(getting and checking the name and give the appropriate response)
+  
+  
+  const constantExpression =`Hello my name is`
+  if(command.startsWith(constantExpression)){
```
Vladimir-malina, line 10: As you did in "calculation - step 8" - you could use in whole function "switch/case" statement
It's more readable and ellegant

javascript/javascript1/week4/homework.js
```diff
@@ -0,0 +1,123 @@
...
+    return `you have ${numberOfToDo}  todo ${namesOfTodo}`
+
+
+   } 
+
+   // part seven(What day is it today)
+   else if (command===`What day is it today?`){
+    let now = new Date()
+    let dateOfNow= now.toDateString()
+    return dateOfNow
```
Vladimir-malina, line 64: Check the formatting of date - it is not as it's expected

javascript/javascript1/week4/homework.js
```diff
@@ -0,0 +1,123 @@
...
+            return result
+   }
+// part nine(Set a timer )
+
+else if(command.startsWith(`Set a timer for`)){
+  const arryOfcommand=command.split(` `);
+  const timer=Number(arryOfcommand[4]);
+  let timeToMs=timer*60000
+  console.log(`Set a timer for ${timer} minutes`)
+  setTimeout (()=>console.log(`Timer done`),timeToMs)
```
Vladimir-malina, line 97: In your test case you add timer for 1 minutes - has to be minute
Try to make solution for both scenarios

javascript/javascript1/week4/homework.js
```diff
@@ -0,0 +1,123 @@
...
+else if(command===`Which exercise is for today?`){
+  const weekday = ["Sunday","Monday","Tuesday","Wednesday","Thursday","Friday","Saturday"];
+  const daysExercise =["fitness","swimming","fitness","walking","fitness","running","cycling",]
+  const days= new Date();
+  const today=days.getDay();
+  return `today is ${weekday[today]} and you should go to ${daysExercise[today]} `
+
+}
+  }
+
```
Vladimir-malina, line 110: Also if you enter new user name - you will still get todos from previous one
Try to reset todos if you get a new user

javascript/javascript1/week4/homework.js
```diff
@@ -0,0 +1,123 @@
...
+
```
Vladimir-malina, line 111: In general solution is working qute well  and we just try to make it a little bit better)
Good job!

PR #116, opened 2024-12-05
javascript-javascript1-week4/jianxin

jianxinz233:
Hand in js1 week4 homework

### DIFF
--- a/javascript/javascript1/week4/voiceAssistant.js
+++ b/javascript/javascript1/week4/voiceAssistant.js
@@ -1,0 +1,183 @@
+let userName = '';
+const toDoList = [];
+
+
+function getReply(command) {
+    command = command.toLowerCase()
+    if(command.startsWith('hello my name is')) {
+        return userNameLogin(command);
+    }
+
+    if (command === 'what is my name') {
+        return replyMyName()
+    }
+
+    if (command.startsWith('add') && command.endsWith('to my todo')) {
+        return addToDoList(command)
+    }
+
+    if (command.startsWith('remove') && command.endsWith('from my todo')) {
+        return removeToDoList(command)
+    }
+
+    if (command === 'what is on my todo?') {
+        return showToDoList()
+    }
+
+    if (command === 'what day is it today?') {
+        return findTodayDate()
+    }
+
+    if (command.startsWith('what is')) {
+        return numberCalculator(command)
+    }
+
+    if (command.startsWith('set a timer for')) {
+        return setTimer(command)
+    }
+    
+    if (command.startsWith('remind me to')) {
+        return setReminder(command)
+    }
+
+    else{
+        return `Sorry, I don't understand your command`
+    }
+}
+
+
+function userNameLogin(command) {
+    let loginName = command.split(' ')[4].toLowerCase().trim();
+    if (loginName === '') {
+        return 'Please tell me your name'
+    }else if (loginName === userName) {
+        return `Nice to meet you again, ${userName}`
+    }else{
+        userName = loginName;
+        return `Nice to meet you ${userName}`
+    }
+}
+
+
+function replyMyName() {
+    if(userName != ''){
+        return `Your name is ${userName}`
+    }else return 'Please tell me your name'
+}
+
+
+function addToDoList(command) {
+    const addStartIndex = command.indexOf('add') + 4;
+    const addEndIndex = command.indexOf('to my todo');
+    let thingToDo = command.slice(addStartIndex, addEndIndex).trim();
+    if (thingToDo === '') {
+        return 'What do you want to add to your todo?'
+    }else if (toDoList.includes(thingToDo)) {
+        return `You have already added ${thingToDo} to your todo`
+    }else{
+        toDoList.push(thingToDo);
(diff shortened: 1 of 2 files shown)


javascript/javascript1/week4/voiceAssistant.js
```diff
@@ -0,0 +1,183 @@
...
+    }
+
+    else{
+        return `Sorry, I don't understand your command`
+    }
+}
+
+
+function userNameLogin(command) {
+    let loginName = command.split(' ')[4].toLowerCase().trim();
```
Vladimir-malina, line 50: It is working good solution,
here you could just use "const" instead of "let" if you are not intended to reassign this variable - it's like common convention "to use as little let as possible - so variable won't get reassign accidently later, which improves code safety and clarity"

Also "toLowerCase()" is not neccessary here as you use this method in "getReply" function for "command" which passed to this function

javascript/javascript1/week4/voiceAssistant.js
```diff
@@ -0,0 +1,183 @@
...
+    if (loginName === '') {
+        return 'Please tell me your name'
+    }else if (loginName === userName) {
+        return `Nice to meet you again, ${userName}`
+    }else{
+        userName = loginName;
+        return `Nice to meet you ${userName}`
+    }
```
Vladimir-malina, line 58: Here function do it's job well - you could just improve code readability avoiding
use of "else if" "else" - conditionals, because you have "return" statements inside and it's will finish function execution and won't let code go into next if statement
Similar as you performed in "getReply" function - but there also "else" isn't necessary but still ok to use

javascript/javascript1/week4/voiceAssistant.js
```diff
@@ -0,0 +1,183 @@
...
+}
+
+
+function replyMyName() {
+    if(userName != ''){
```
Vladimir-malina, line 63: I would recommend always to use strict equality and inequality operators, unless you have a specific reason for not doing that
 '!==' insted of '!='
 Try 0 != '' or false!='', what are you going to get here?

javascript/javascript1/week4/voiceAssistant.js
```diff
@@ -0,0 +1,183 @@
...
+
+function userNameLogin(command) {
+    let loginName = command.split(' ')[4].toLowerCase().trim();
+    if (loginName === '') {
+        return 'Please tell me your name'
+    }else if (loginName === userName) {
+        return `Nice to meet you again, ${userName}`
+    }else{
+        userName = loginName;
+        return `Nice to meet you ${userName}`
```
Vladimir-malina, line 57: Also when you get a new user would be nice to reset todo list to empty array, in order not to have a saved todos for another user.
We don't  want to go deep -  to save users and appropriate todos for each user and when familiar user "returns" have it's todos ready, but just to reset all todos would be "ok" logic

javascript/javascript1/week4/voiceAssistant.js
```diff
@@ -0,0 +1,183 @@
...
+    let todayDateDay= todayDate.getDate();
+
+    return `Today is ${todayDateDay}. of ${todayDateMonth} ${todayDateYear}` 
+}
+
+
+function numberCalculator(command) {
+    let calculatedString = command.slice(8).trim();
+    let result = eval(calculatedString);
+    return result;
```
Vladimir-malina, line 127: "eval"  is easy and working approach but it's generally not recommended to use due to potential code injection attacks as eval() executes any code passed to it

javascript/javascript1/week4/voiceAssistant.js
```diff
@@ -0,0 +1,183 @@
+let userName = '';
+const toDoList = [];
+
+
```
Vladimir-malina, line 4: Really good decision to extract logic of each conditional into different smaller functions - so code stayed easy to read and furthermore to maintain and debug
Good job!

javascript/javascript1/week4/voiceAssistant.js
```diff
@@ -0,0 +1,183 @@
...
+    }
+}
+
+
+function findTodayDate() {
+    let todayDate = new Date();
+    let todayDateYear = todayDate.getFullYear();
+    const allMonths = ['January', 'Feburary', 'March', 'April', 'May', 'June', 'July', 'August', 'September', 'October', 'November', 'December']
+    let todayDateMonth = allMonths[todayDate.getMonth()]
+    let todayDateDay= todayDate.getDate();
```
Vladimir-malina, line 118: Good solution here and if you don't want to create an array with all months
you could use inbuilt method - todayDate.toLocaleString('default', { month: 'long' });
in order to get a full name of the month

javascript/javascript1/week4/voiceAssistant.js
Vladimir-malina: The same with "else if", and "else" troughout the whole code

javascript/javascript1/week4/voiceAssistant.js
Vladimir-malina: Same with "let" and "const" throughout the whole code

PR #119, opened 2024-12-05
VoiceAssistent execise

Aymanngu:


javascript/javascript1/week4/VoiceAssistent.js
```diff
@@ -0,0 +1,111 @@
...
+    if (userName === "") {
+      return "I don't know your name yet.";
+    } else {
+      return `Your name is ${userName}`;
+    }
+  }
+
+  // Todos Part
+  if (command.startsWith("Add") && command.endsWith("to my todo")) {
+    const task = command.replace("Add ", "").replace(" to my todo", "");
```
jmf-umbraco, line 27: What happens if the thing that the user wants to add to their todo list includes the text "Add "?

javascript/javascript1/week4/VoiceAssistent.js
```diff
@@ -0,0 +1,111 @@
...
+    todos.push(task);
+    return `${task} added to your todo`;
+  }
+
+  if (command.startsWith("Remove") && command.endsWith("from my todo")) {
+    const task = command.replace("Remove ", "").replace(" from my todo", "");
+    const index = todos.indexOf(task);
+    if (index > -1) {
```
jmf-umbraco, line 35: Good edge case handling!

javascript/javascript1/week4/VoiceAssistent.js
```diff
@@ -0,0 +1,111 @@
...
+    } else {
+      return `You have ${todos.length} todos: ${todos.join(", ")}`;
+    }
+  }
+
+  // Day Part
+  if (command === "What day is today?") {
+    const today = new Date();
+    const day = today.getDate();
+    const month = today.toLocaleString("en-US", { month: "long" });
```
jmf-umbraco, line 55: Good use of `.toLocaleString()`, but we might be able to save ourselves some further effort by using it for the day and year as well!

Also, for real apps we should consider whether `en-US` is the correct choice of culture. If the rest of your app is in English, then perhaps just `en` without a country code is preferable. If your app is localised such that users see their native language anywhere, then we might want to use `default` so that the date doesn't appear in English!

javascript/javascript1/week4/VoiceAssistent.js
```diff
@@ -0,0 +1,111 @@
...
+  }
+
+
+if (command.startsWith("What is")) {
+    let [num1, operator, num2] = command.replace("What is ", "").split(" ");
+    num1 = parseFloat(num1);
+    num2 = parseFloat(num2);
+  
+    if (isNaN(num1) || isNaN(num2)) {
+      return "I couldn't understand the numbers.";
```
jmf-umbraco, line 67: Consider using an error message that gives the user more clues about how they should proceed. For example, you could help the user understand how they should be formatted, or let them know that perhaps they're missing some whitespace between the numbers and operators.

javascript/javascript1/week4/VoiceAssistent.js
```diff
@@ -0,0 +1,111 @@
+let userName = ""; 
+const todos = []; 
+
+function getReply(command) {
+  // Name Part
+  if (command.startsWith("Hello my name is")) {
```
jmf-umbraco, line 6: What happens if the user types the command in all lower case? Would be nice if our code just handled such a situation!

jmf-umbraco (changes requested):
Great job overall. Tidy code and some clever ways of solving the various tasks.

I've left a few small suggestions you may want to follow-up on to make the code a little more robust or readable!

PR #121, opened 2024-12-06
Javascript javascript1 week4/ruslana

ruslanaonshchk:
Did all tasks from the class (01.12) as I was absent and did the main voice assistant task

### DIFF
--- a/javascript/javascript1/week4/script.js
+++ b/javascript/javascript1/week4/script.js
@@ -1,0 +1,98 @@
+let userName = null; 
+let todoList = []; 
+
+function getReply(command) {
+  const lowerCaseCommand = command.toLowerCase();
+
+  if (lowerCaseCommand.startsWith('hello my name is ')) {
+    const name = command.slice(17); 
+    if (userName === name) {
+      return `You already introduced yourself as ${name}.`;
+    } else {
+      userName = name;
+      return `Nice to meet you ${name}`;
+    }
+  }
+
+  if (lowerCaseCommand === 'what is my name?') {
+    return userName
+      ? `Your name is ${userName}`
+      : "I don't know your name yet. Please tell me your name first.";
+  }
+
+  if (
+    lowerCaseCommand.startsWith('add ') &&
+    lowerCaseCommand.includes(' to my todo')
+  ) {
+    const todo = command.slice(4, command.indexOf(' to my todo'));
+    todoList.push(todo);
+    return `${todo} added to your todo.`;
+  }
+
+  if (
+    lowerCaseCommand.startsWith('remove ') &&
+    lowerCaseCommand.includes(' from my todo')
+  ) {
+    const todo = command.slice(7, command.indexOf(' from my todo'));
+    const index = todoList.indexOf(todo);
+    if (index > -1) {
+      todoList.splice(index, 1); 
+      return `Removed ${todo} from your todo.`;
+    } else {
+      return `${todo} is not in your todo list.`;
+    }
+  }
+
+  if (lowerCaseCommand === 'what is on my todo?') {
+    if (todoList.length === 0) {
+      return 'Your todo list is empty.';
+    } else {
+      return `You have ${todoList.length} todos: ${todoList.join(', ')}.`;
+    }
+  }
+
+  if (lowerCaseCommand === 'what day is it today?') {
+    const today = new Date();
+    const day = today.getDate();
+    const month = today.toLocaleString('default', { month: 'long' }); 
+    const year = today.getFullYear();
+    return `${day}. of ${month} ${year}`;
+  }
+
+  if (lowerCaseCommand.startsWith('what is ')) {
+    const mathExpression = command.slice(8); 
+    try {
+      const result = eval(mathExpression); 
+      return `${mathExpression} is ${result}`;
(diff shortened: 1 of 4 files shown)


javascript/javascript1/week4/script.js
```diff
@@ -0,0 +1,98 @@
+let userName = null; 
+let todoList = []; 
+
+function getReply(command) {
+  const lowerCaseCommand = command.toLowerCase();
```
jmf-umbraco, line 5: Good thinking! It's always nice to make an effort to handle slightly incorrect user input, such as casing.

javascript/javascript1/week4/script.js
```diff
@@ -0,0 +1,98 @@
...
+    return userName
+      ? `Your name is ${userName}`
+      : "I don't know your name yet. Please tell me your name first.";
+  }
+
+  if (
+    lowerCaseCommand.startsWith('add ') &&
+    lowerCaseCommand.includes(' to my todo')
+  ) {
+    const todo = command.slice(4, command.indexOf(' to my todo'));
```
jmf-umbraco, line 27: It's not something super important to handle, but it's always good to think about edge cases and how our code might behave.

For example, what if someone wanted to add `think of some things to add to my todo` to their todo list? Their command would be:
```
add think of some things to add to my todo to my todo
```
Which is a little silly for sure! But it would be nice if our code handled it.

What the code will currently do is see the first instance of `to my todo` when we call `command.indexOf("to my todo")`. That means it will assign `think of some things to add` to the `todo` variable, whereas it would ideally assign `think of some things to add to my todo` instead.

We can fix this by using a slightly different method from `indexOf()`. Fortunately Javascript has a `lastIndexOf()` method for exactly this reason. Instead of returning the index of the first matching string, it will return the index of the last one!

Again - not a super important fix for now. The lesson is mostly just to keep thinking about our code and how it might break, so that we can make deliberate choices about whether we fix it!

javascript/javascript1/week4/script.js
```diff
@@ -0,0 +1,98 @@
...
+    return `${todo} added to your todo.`;
+  }
+
+  if (
+    lowerCaseCommand.startsWith('remove ') &&
+    lowerCaseCommand.includes(' from my todo')
+  ) {
+    const todo = command.slice(7, command.indexOf(' from my todo'));
+    const index = todoList.indexOf(todo);
+    if (index > -1) {
```
jmf-umbraco, line 38: Great catch. Calling `.splice()` with `-1` as an argument would have very unexpected results!

javascript/javascript1/week4/script.js
```diff
@@ -0,0 +1,98 @@
...
+      return 'Your todo list is empty.';
+    } else {
+      return `You have ${todoList.length} todos: ${todoList.join(', ')}.`;
+    }
+  }
+
+  if (lowerCaseCommand === 'what day is it today?') {
+    const today = new Date();
+    const day = today.getDate();
+    const month = today.toLocaleString('default', { month: 'long' });
```
jmf-umbraco, line 57: Great use of some newer Javascript functionality - `toLocaleString()`. Consider using it for the whole date format though, not just the month!

javascript/javascript1/week4/script.js
```diff
@@ -0,0 +1,98 @@
...
+    const day = today.getDate();
+    const month = today.toLocaleString('default', { month: 'long' }); 
+    const year = today.getFullYear();
+    return `${day}. of ${month} ${year}`;
+  }
+
+  if (lowerCaseCommand.startsWith('what is ')) {
+    const mathExpression = command.slice(8); 
+    try {
+      const result = eval(mathExpression);
```
jmf-umbraco, line 65: While `eval()` is very useful for cases like this where we want to evaluate an expression, we need to be very careful about using it as developers.

Many sites will disable using it entirely these days, but consider what happens if the user types
```
what is alert(`hello ${userName}`)
```

That's right - this user input will cause the browser (if running in the browser) to pop up an alert dialog, and it's even got access to the value of any variables! Consider that this could be used to send information about the user to a malicious site, and you'll start to realise why `eval()` is generally considered best left alone!

That said, for this assignment, it's a great way to get started.

An alternative could be to parse some numbers and an operator (+, -, *, /) form the command string, and then perform a different operation based on the operator.

javascript/javascript1/week4/script.js
```diff
@@ -0,0 +1,98 @@
...
+      return "Sorry, I couldn't calculate that.";
+    }
+  }
+
+  if (
+    lowerCaseCommand.startsWith('set a timer for ') &&
+    lowerCaseCommand.includes(' minutes')
+  ) {
+    const minutes = parseInt(command.slice(16, command.indexOf(' minutes')));
+    if (isNaN(minutes)) {
```
jmf-umbraco, line 77: Good error handling!

jmf-umbraco (changes requested):
Overall, looks good. I would suggest breaking your `getReply()` function up into multiple - it's common to have a "main" function that delegates different tasks to helper functions.

So for example, you could have functions like `addTodo` or `setUserName` etc. and call them from the `getReply()` function. This could help keep the length of the `getReply()` function down and help keep it readable at-a-glance.

PR #122, opened 2024-12-06
Javascript javascript1 week4/guzide

GuzideGuzelbey:
Thanks for the challenging exercise, huh. 

Moved the week4 practice to a new branch from main which was mistakenly branched out from week3.
So I am creating the request again :)

### DIFF
--- a/javascript/javascript1/week4/week4-guzide/voiceAssistant.js
+++ b/javascript/javascript1/week4/week4-guzide/voiceAssistant.js
@@ -1,0 +1,153 @@
+const toDo = [];
+let userName = "";
+
+function getReply(command) {
+  if (command.startsWith("Hello my name is")) {
+    return setUserName(command);
+  }
+
+  if (command.startsWith("What is my name?")) {
+    return getUserName();
+  }
+
+  if (command.startsWith("Add")) {
+    return addToDo(command);
+  }
+
+  if (command.startsWith("Remove")) {
+    return removeToDo(command);
+  }
+
+  if (command === "What is on my todo?") {
+    return showToDos();
+  }
+
+  if (command === "What day is it today?") {
+    return getAndFormatDate();
+  }
+
+  if (command.startsWith("what is")) {
+    return calculate(command);
+  }
+
+  if (command.startsWith("Set a timer for")) {
+    return setTimer(command);
+  }
+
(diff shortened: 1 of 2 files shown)


javascript/javascript1/week4/week4-guzide/voiceAssistant.js
```diff
@@ -0,0 +1,153 @@
...
+        return b !== 0 ? a / b : "Cannot divide by zero";
+      default:
+        return "Sorry, this operator is not supported.";
+    }
+  }
+  return "Error! Invalid calculation command.";
+}
+
+function setTimer(input) {
+  const time = input.match(/Set a timer for (\d+) minutes/i);
```
Khandohii: What if to set timer for "1 minute". I would add a validation for "minute" as well.

javascript/javascript1/week4/week4-guzide/voiceAssistant.js
```diff
@@ -0,0 +1,153 @@
...
+      case "*":
+        return a * b;
+      case "/":
+        return b !== 0 ? a / b : "Cannot divide by zero";
+      default:
+        return "Sorry, this operator is not supported.";
+    }
+  }
+  return "Error! Invalid calculation command.";
+}
```
Khandohii, line 123: Good solution. For solving mathematical calculations you can use `eval()` function as well.

Khandohii (approved):
Good job @GuzideGuzelbey 👍 
Left some comments 😊

javascript/javascript1/week4/week4-guzide/voiceAssistant.js
GuzideGuzelbey: Hi Khandohii, thank you. But while I was searching for the methods to be used I came up with this info "eval() executes any JavaScript code in the input string. This makes it vulnerable to injection attacks." That is why I did not use it. What do you think about this info?

javascript/javascript1/week4/week4-guzide/voiceAssistant.js
GuzideGuzelbey: Validation is added, thank you :)

GuzideGuzelbey:
Overall, thanks for the review and the constructive feedbacks.

PR #123, opened 2024-12-06
Javascript javascript1 week4/andriiK

pahanan:


javascript/javascript1/week4/Highest_and_Lowest.js
```diff
@@ -0,0 +1,32 @@
+function highestandLowest(stringOfNumbers){
+    const array = stringOfNumbers.split(" ").map(Number);
+
+    const numbers = ["1234567890 "];
```
AngelHenriettaAboah, line 4: const numbers = [" 123456789 "]; is declared but not used anywhere. can we get rid of it.

javascript/javascript1/week4/Square_Every_Digit.js
```diff
@@ -0,0 +1,26 @@
+function squareEveryDigit(number){
+    if(typeof(number) !== "number"){
```
AngelHenriettaAboah, line 2: would be nice to add a check for isNaN(number).

javascript/javascript1/week4/Square_Every_Digit.js
```diff
@@ -0,0 +1,26 @@
...
+    }
+    
+    const digits = number.toString();
+    let result = "";
+    
+    for(let digit of digits){
+        result += (digit*digit).toString();
+    }
+
+    result = parseInt(result, 10)
```
AngelHenriettaAboah, line 13: Consider returning the string directly since the expected output is a string and the squaring result doesn't require  leading zeros.

javascript/javascript1/week4/Voice_assistant.js
```diff
@@ -0,0 +1,101 @@
...
+        const today = new Date();
+        const day = today.getDate();
+        const month = today.toLocaleString("default", { month: "long" });
+        const year = today.getFullYear();
+        return `Today is ${day}. of ${month} ${year}`;
+    }
+
+    if (command.startsWith("What is")) {
+        const mathExpression = command.slice(8).trim();
+        const result = eval(mathExpression);
```
AngelHenriettaAboah, line 74: Replace eval with a safer alternative. For example, you can use the Function constructor in strict mode to evaluate expressions securely.

javascript/javascript1/week4/Vowel_Count.js
```diff
@@ -0,0 +1,26 @@
+function countVowel(givenString){
+    if(givenString === "" || typeof(givenString) !== "string"){
+        return "error";
+    } 
+
+    let vowelCount = 0;
+    const vowels = ["a", "e", "i", "o", "u"];
+
+    for (let vowel of vowels) {
+        for (let letter of givenString) {
```
AngelHenriettaAboah, line 10: You only need to loop through the "givenString" once and check if each character is a vowel using "vowels.includes(letter)."

AngelHenriettaAboah (changes requested):
Good job Andrii.. I left a few comments!

PR #126, opened 2024-12-10
Javascript javascript1 week4/soheib

siderdk:
Dear Mentor, 
First of all, please accept my sincere apologies for the late submission. Things should get back on track, and I will be submitting on time. 
I have included in this assignment many comments. I apologize if the interfere with the readability of the code. It is just to assist me when I come back to the assignment (which I know I will). 
Thank you again.

### DIFF
--- a/javascript/javascript1/week4/homework6/scripts.js
+++ b/javascript/javascript1/week4/homework6/scripts.js
@@ -1,0 +1,1 @@
+function
--- a/javascript/javascript1/week4/homework6/scripts.js
+++ b/javascript/javascript1/week4/homework6/scripts.js
@@ -1,1 +1,133 @@
-function
+let todos = []; // initializing an empty list of todo
+let loggedName = []; //initializing an empty list of names
+
+//function to capture the name from the command
+function getName(command){
+    const name = command.split(' ')[4]; // note to self, split is a string method, it takes a separtor argument, in this case it's a space, and returns an array populated with the separeted elements as strings, I am accessing the 5th element directly
+    return name
+}
+
+//function to capture a task, either to add or remove
+function getTask(command) {
+    if(command.startsWith('Add') || command.startsWith('add')){
+    const endIndex = command.indexOf(' to my todo') // I plan on using slice later, so I'm getting the index at which the slice should stop
+    const task = command.slice(4, endIndex) // get the part of the string between "Add" and "to my todo"
+    return task}
+    //do the same thing except adjust for the words "from" and "Remove"
+    else if (command.startsWith('Remove') || command.startsWith('remove')){
+    const endIndex = command.indexOf(' from my todo') 
+    const task = command.slice(7, endIndex) 
+    return task
+    }
+}
+
+//function to remove a task from the todos
+function removeTask(task) {
+    const taskIndex = todos.findIndex(task);
+    todos.slice(taskIndex, 1)
+    return console.log(`Removed ${task} from your todo`)
+}
+
+// function print date in human redable format
+function getDate(){
+    const today = new Date();
+    const month = today.getMonth(); 
+    const months = ["January", "February", "March", "April", "May", "June", "July", "August", "September", "October", "November", "December"];
+    return console.log(`Today is the ${today.getDate()}. of ${months[month]} ${today.getFullYear()}`)
+}
+
+// function for timer 
+function timer(command) {
+    function getInterval(command){
+        const regex = /(\d+)/ //regular expression to find digits in the command str
+        const match = command.match(regex) // this method will return an array with matching values to the regex 
+        return match ? Number(match[0]) : null; // if true convert the regex to a number and return it
+    }
+    const number = getInterval(command)
+    const interval = number * 60 * 1000;
+    function logDone(){
+        console.log(`Timer Done`)
+    }
+    setTimeout(logDone, interval)
+    console.log(`A timer for ${number} minute(s) has been set`)
+}
+
+
+
+//main function to get answers from the "voice assistant"
+function getReply(command) {
+    // here we deal with the greeting commands
+    if (command.startsWith('Hello my name is')){
+        const person = getName(command)
+        if (!loggedName.includes(person)) {
+            loggedName.push(person)
+            return console.log(`Nice to meet you ${person}`)
+        } else if (loggedName.includes(person)) {
+            return console.log(`Hey ${person}! we met already.`)
+        }
+    }
+    // here we deal with reminding the user their name
+    else if (command.startsWith('What is my name')) {
+        if (loggedName.length === 0) {
+            return console.log(`I can't find a name. \n Please introduce yourself in this way: "Hello my name is" followed by your name`)
+        } else if (loggedName.length > 0) {
+            const recentName = loggedName[loggedName.length -1]
+            return console.log(`Your name is ${recentName}! What can I help you with, ${recentName}?`)
+        }
+    }
+    //dealing with the todo part
+    else if (command.includes('my todo')) {
+        let newTask = getTask(command);
+        //adding a task
+        if(command.startsWith('Add') || command.startsWith('add')){
+           todos.push(newTask); 
+           return console.log(`${newTask} has been added to your todos`)
+        
+        } 
+        //removing a task
+        else if (command.startsWith('Remove') || command.startsWith('remove')){
+            const taskIndex = todos.indexOf(newTask);
+            todos.splice(taskIndex, 1)
+            return console.log(`${newTask} has been removed from your todos`)
+        }
+        // showing all task 
+        else if (command.startsWith('What is on my todo?')){
+            if (todos.length === 0) {
+                console.log(`Your list is empty, try adding some tasks first`)
+            } else {
+                console.log(`Your todo list is comprised of: \n${todos.join('\n')}`)
+            }
+        }
+    }
+    // dealing with the date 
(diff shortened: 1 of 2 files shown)


javascript/javascript1/week4/homework6/scripts.js
```diff
@@ -0,0 +1,133 @@
+let todos = []; // initializing an empty list of todo
+let loggedName = []; //initializing an empty list of names
```
jmf-umbraco, line 2: Consider whether this should be an array.

We seem to be doing a lot of work later on (lines 60~75) to account for the fact that we're storing old names in an array. It might be easier if we simply store the most recent name that the user has provided as a single string!

javascript/javascript1/week4/homework6/scripts.js
```diff
@@ -0,0 +1,133 @@
...
+
+//function to capture the name from the command
+function getName(command){
```
jmf-umbraco, line 5: Nice one splitting up the code into several functions with clear purposes.

javascript/javascript1/week4/homework6/scripts.js
```diff
@@ -0,0 +1,133 @@
...
+    const name = command.split(' ')[4]; // note to self, split is a string method, it takes a separtor argument, in this case it's a space, and returns an array populated with the separeted elements as strings, I am accessing the 5th element directly
```
jmf-umbraco, line 6: This is quite minor, but I'll explain my thinking so you can decide for yourself when writing future code!

When we create variables it should _usually_ be one of several reasons:
* We want to re-use it more than once, so referring to it by a nice short repeatable name is useful
* It would be hard to read if we put the code to calculate the variable in-line instead
* We would like to give something complex a nice variable name so that we can have readable, self-documenting code

In this case, it might actually just create cleaner code to drop the `const` variable declaration, and simply `return command.split(' ')[4];`

javascript/javascript1/week4/homework6/scripts.js
```diff
@@ -0,0 +1,133 @@
...
+    return name
+}
+
+//function to capture a task, either to add or remove
+function getTask(command) {
+    if(command.startsWith('Add') || command.startsWith('add')){
```
jmf-umbraco, line 12: Good job handling case sensitivity.

However, there could be other ways you might want to handle it - perhaps you could consider using `.toUpperCase()` or `.toLowerCase()` to normalize the string before your `if (...)` statement 😀

javascript/javascript1/week4/homework6/scripts.js
```diff
@@ -0,0 +1,133 @@
...
+        else if (command.startsWith('What is on my todo?')){
+            if (todos.length === 0) {
+                console.log(`Your list is empty, try adding some tasks first`)
+            } else {
+                console.log(`Your todo list is comprised of: \n${todos.join('\n')}`)
+            }
+        }
+    }
+    // dealing with the date 
+    else if (command.startsWith('What day is it today?') || command.startsWith('What day is it') ) {
```
jmf-umbraco, line 103: The second part of this statement will always be true if the first half is, so we may wish to drop the first part entirely!

javascript/javascript1/week4/homework6/scripts.js
```diff
@@ -0,0 +1,133 @@
...
+function getDate(){
+    const today = new Date();
+    const month = today.getMonth(); 
+    const months = ["January", "February", "March", "April", "May", "June", "July", "August", "September", "October", "November", "December"];
+    return console.log(`Today is the ${today.getDate()}. of ${months[month]} ${today.getFullYear()}`)
+}
+
+// function for timer 
+function timer(command) {
+    function getInterval(command){
```
jmf-umbraco, line 41: Good stuff using helper functions to do things like this. We do generally want to be careful about nesting functions though - too many levels and our code can become difficult to read!

javascript/javascript1/week4/homework6/scripts.js
```diff
@@ -0,0 +1,133 @@
...
+            loggedName.push(person)
+            return console.log(`Nice to meet you ${person}`)
+        } else if (loggedName.includes(person)) {
+            return console.log(`Hey ${person}! we met already.`)
+        }
+    }
+    // here we deal with reminding the user their name
+    else if (command.startsWith('What is my name')) {
+        if (loggedName.length === 0) {
+            return console.log(`I can't find a name. \n Please introduce yourself in this way: "Hello my name is" followed by your name`)
```
jmf-umbraco, line 72: Fantastic error message!

As developers, we are frequently guilty of writing error messages that say things like "Invalid input", or some variation on "I can't do that". This can be frustrating to the user - because from their perspective they have no idea what's actually wrong, or what else they can try.

Your error message instead helps the user to understand how the program does work and gives clues about what they can do about the error. It's much more helpful!

javascript/javascript1/week4/homework6/scripts.js
```diff
@@ -0,0 +1,133 @@
...
+    // dealing with math operations
+    else if ( command.startsWith('What is') && (/[+\-*/]/.test(command))){ //apparently .includes does not work with regex, so instead you se .test
+        const operation = command.slice(7)
+        return eval(operation)
+    }
+    // setting a timer 
+    else if ( command.startsWith('Set a timer')) {
+        return timer(command)
+    }
+    //additional feature
```
jmf-umbraco, line 115: 😂

javascript/javascript1/week4/homework6/scripts.js
```diff
@@ -0,0 +1,133 @@
...
+
+//function to remove a task from the todos
+function removeTask(task) {
+    const taskIndex = todos.findIndex(task);
+    todos.slice(taskIndex, 1)
+    return console.log(`Removed ${task} from your todo`)
+}
+
+// function print date in human redable format
+function getDate(){
```
jmf-umbraco, line 32: There's a neat (newish) method in Javascript called `.toLocaleString()` that will format a date based on some options you provide to it. It could be easier than doing the formatting this way.

For example:
```
let date = new Date();
let dateFormatOptions = {
    month: "long",
    year: "numeric"
};

```

javascript/javascript1/week4/homework6/scripts.js
```diff
@@ -0,0 +1,133 @@
...
+        }
+    }
+    // dealing with the date 
+    else if (command.startsWith('What day is it today?') || command.startsWith('What day is it') ) {
+       return getDate()
+    }
+    // dealing with math operations
+    else if ( command.startsWith('What is') && (/[+\-*/]/.test(command))){ //apparently .includes does not work with regex, so instead you se .test
+        const operation = command.slice(7)
+        return eval(operation)
```
jmf-umbraco, line 109: While `eval()` is very useful for cases like this where we want to evaluate an expression, we need to be very careful about using it as developers.

Many sites will disable using it entirely these days, but consider what happens if the user types
```
what is alert(`hello ${loggedName[-1]}`)
```

That's right - this user input will cause the browser (if running in the browser) to pop up an alert dialog, and it's even got access to the value of any variables! Consider that this could be used to send information about the user to a malicious site, and you'll start to realise why `eval()` is generally considered best left alone!

That said, for this assignment, it's a great way to get started.

An alternative could be to parse some numbers and an operator (+, -, *, /) form the command string, and then perform a different operation based on the operator.

jmf-umbraco (changes requested):
Overall looks good! I've left a few suggestions of things you may want to change to a slightly simpler or more robust alternative 😀

javascript/javascript1/week4/homework6/scripts.js
siderdk: Thank you for the valuable feedback.  
My first implementation was with a string, however, I thought it would give a better "experience" if I stored all entered names, instead of only the latest.

javascript/javascript1/week4/homework6/scripts.js
siderdk: Duly noted. 
It was an oversight from my side. 
Thank you

javascript/javascript1/week4/homework6/scripts.js
siderdk: That is a better approach. Thank you!

javascript/javascript1/week4/homework6/scripts.js
siderdk: Another oversight. Thank you for pointing out the redundancy🙏

javascript/javascript1/week4/homework6/scripts.js
siderdk: Is it a good rule to go then: "I will not nest functions ever. Instead I'll define them globally and then just call them when needed". ? 
Thank you again

javascript/javascript1/week4/homework6/scripts.js
siderdk: Many thanks 🙏

javascript/javascript1/week4/homework6/scripts.js
siderdk: Woooooaaaaaaw 
That is awesome knowledge!!
Thank you so much. I will try the parsin method

siderdk:
> Dear Mentor, First of all, please accept my sincere apologies for the late submission. Things should get back on track, and I will be submitting on time. I have included in this assignment many comments. I apologize if the interfere with the readability of the code. It is just to assist me when I come back to the assignment (which I know I will). Thank you again.

Thank you so much for the very kind and valuable feedback. 
Really insightful. I will adopt the suggested changes

PR #128, opened 2024-12-12
Javascript javascript2 week1 aminata

Aminata00:
homework--
my understanding of the assignment is to make an app where the user will be able to create his own quiz, and then reply to the questions. So this the first part to create it. 

I have had some problems regarding how to addEventListener to the save options buttons. If you have some tips it'll be greatly appreciated :))!

### DIFF
--- a/javascript/javascript2/week1/quizzapp.html
+++ b/javascript/javascript2/week1/quizzapp.html
@@ -1,0 +1,35 @@
+<!DOCTYPE html>
+<body>
+    <header></header>
+
+    <section>
+        <label>Add a question
+            <input id="add-question" type="text" />
+            <button id="add-question-button">Add to the quiz</button>
+        </label>
+    </section>
+
+    <section id="options-section">
+        <label>Type possible answers:</label>
+        <label>Option 1:
+            <input id="option1" type="text" class="option-input" />
+            <button class="save-option" data-option="option1">Save</button>
+        </label>
+        <label>Option 2:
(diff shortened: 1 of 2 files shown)


LucasIversen (changes requested):
Hi Aminata.

I think you have to take a look at the project description again. A quiz object has to look like this:

`const quizQuestion = {
  id: 1,
  question: "What is the capital of Denmark?",
  options: [
    { text: "Berlin", isCorrect: false },
    { text: "Copenhagen", isCorrect: true },
    { text: "Madrid", isCorrect: false },
    { text: "Rome", isCorrect: false },
  ],
  explanation: "Copenhagen is the capital of Denmark.",
};`

So we need to make a form to create these quiz items :)

Let me know if more needs explaining

Aminata00:
> Hi Aminata.
> 
> I think you have to take a look at the project description again. A quiz object has to look like this:
> 
> `const quizQuestion = { id: 1, question: "What is the capital of Denmark?", options: [ { text: "Berlin", isCorrect: false }, { text: "Copenhagen", isCorrect: true }, { text: "Madrid", isCorrect: false }, { text: "Rome", isCorrect: false }, ], explanation: "Copenhagen is the capital of Denmark.", };`
> 
> So we need to make a form to create these quiz items :)
> 
> Let me know if more needs explaining

oh my god so I completely misunderstood right ? 😂 
Thank you I'll make the necessary changes !!

PR #141, opened 2025-02-02
Quizz app

StefanyRomeroV:
Problems with the timer and the page game, where it doesn't display the corrects and incorrect - working on that- next is to do it with the API

### DIFF
--- a/javascript/javascript3/week3/game.css
+++ b/javascript/javascript3/week3/game.css
@@ -1,0 +1,216 @@
+* {
+    margin: 0 ;
+    padding: 0;
+    font-family: 'Poppins', sans-serif;
+}
+
+
+body {
+    display: flex; 
+    justify-content: center; 
+    align-items: center; 
+    height: 100vh; 
+    margin: 0; 
+    background-color: #f0f0f0; 
+}
+
+.panel-add {
+    width: 80%;
+    height: 80%;
+    max-width: 800px;
+    min-height: 40vh;
+    background-color: #fff;
+    margin: 0; 
+    border-radius: 10px;
+    padding: 30px;
+    border: 1px solid #eb8328;
+    box-shadow: 0 4px 6px rgba(0, 0, 0, 0.1);
+}
+.panel-add h1 {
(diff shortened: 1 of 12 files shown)


javascript/javascript3/week3/game.css
```diff
@@ -0,0 +1,216 @@
...
+        width: 95%;
+        padding: 20px;
+
+    
+    }
+
+
+    .players-name input {
+        width: 80%;
+        ;
```
te-online, line 198: Looks like you've left an extra `;` here 😊 🗑️

javascript/javascript3/week3/game.css
```diff
@@ -0,0 +1,216 @@
...
+}
+
+.incorrect {
+    background: rgb(239, 174, 174);
+
+}
+
+
+/* Responsive adjustments */
+@media (max-width: 768px) {
```
te-online, line 187: In general, try to format your CSS file, it will help readability 💅

te-online (commented):
Generally, I'd like to approve this PR ✅ 

⚠️ However, I'm not sure if I should approve here in Github, because the code should _not_ be merged to this repository. You were supposed to create new repositories for your JavaScript projects. This way you can showcase your projects and also deploy them to, e.g. render.com.

Now some feedback to your app 😊 

- 👍 I like the general vibe of the app (started from `index.html`), giving you just enough time to answer the questions and keeping the momentum by sending you to the next question. There's just one small bug 🐛 ; when you finish the quiz, the timer should stop. Now it just runs out and the "restart" button somehow disappears.
- 💅 In regards to your code, you could try to format and comment your files (especially JS and CSS), which makes them much more readable. It makes it easier for other people to understand what is going on and easier for you to come back to it later.
- ♻️ In general, I like the idea of splitting different aspects into different pages (your quiz, general quiz). But you could try to split your code, so that all pages would share the things you write over and over again. This may be a bit advanced and copying working functionality is always an option, but it could be a next step to reduce duplicate code between your pages/files.

PR #144, opened 2025-02-04
Database week1/gayathri

GayathriVenkatraman:


databases/week1/homework.sql
```diff
@@ -0,0 +1,44 @@
+-- Find out how many tasks are in the task table
+SELECT COUNT(*) as total_tasks FROM task
+
+-- Find out how many tasks in the task table do not have a valid due date
+SELECT * from task WHERE due_date is NULL
```
josejfernandez: This returns information about the tasks, but does not count them. Please rewrite the query so it returns the number.

databases/week1/homework.sql
```diff
@@ -0,0 +1,44 @@
...
+
+-- Find all the tasks that are marked as done
+SELECT * FROM task WHERE status_id IN (SELECT id from status where name LIKE "%DONE%")
```
josejfernandez: This does not block the review because your query works, but I still suggest you follow this advice to get more clear and meaningful code.

You use `IN` when there is a list of possible values. Although it works in this case, there is only one status for "Done", so you could do `SELECT * FROM task WHERE status_id = (...)` and this statement would be a bit more clear.

Similarly, in the subquery you are searching for statuses whose name contain `DONE`. There is only one status for "Done", so you could do `SELECT id from status where name = 'Done'`, which again is a bit more clear.

databases/week1/homework.sql
```diff
@@ -0,0 +1,44 @@
...
+
+-- Find all the tasks that are not marked as done
+SELECT * FROM task WHERE status_id not IN (SELECT id from status where name LIKE "%DONE%")
+
+-- Get all the tasks, sorted with the most recently created first
+SELECT * from task ORDER BY created DESC
```
josejfernandez: I have noticed inconsistent casing in the different statements: `FROM` in the first one, `from` in the second, then there is also `as` and `AS`. Although this is not blocking the review, consider sticking to a particular way of writing code :)

My suggestion is, that you type SQL keywords in uppercase: `SELECT COUNT(id) FROM task;`

josejfernandez (changes requested):
Hey Gayathri! Most of your queries worked flawlessly, good job there! There is one query that is not returning exactly what is being requested. Please fix that and I can approve your pull request. The other two comments are not blocking, so I suggest you address them but it's up to you.

databases/week1/homework.sql
GayathriVenkatraman: Sorry, it's my mistake and it is corrected now.

databases/week1/homework.sql
GayathriVenkatraman: Okay, thanks for your advice modified the code based on your comments.

databases/week1/homework.sql
GayathriVenkatraman: Yes I just noticed and changed it. It was because sometimes I used the intellicence and sometimes I ignored it, so that has caused the inconsistent casing.

GayathriVenkatraman:
Hi Jose,

Thank you for your time on reviewing the code and valuable feedback. 

Regards,
Gayathri

PR #145, opened 2025-02-04
databases week1/farzaneh

FarzanehAhmadi:


databases/week1/Homework.sql
```diff
@@ -0,0 +1,49 @@
+-- 1. Find out how many tasks are in the task table
+Select count(id) from task;
+-- Result: 35
+
+-- 2. Find out how many tasks in the task table do not have a valid due date
+select COUNT(id) from task
```
josejfernandez: I have noticed inconsistent casing in the different statements: `Select` in the first one, `select` in the second, `SELECT` in the third. Although this is not blocking the review, consider sticking to a particular way of writing code :)

My suggestion is, that you type SQL keywords in uppercase: `SELECT COUNT(id) FROM task;`

databases/week1/Homework.sql
```diff
@@ -0,0 +1,49 @@
...
+
+-- 8. Get the title and status (as text) of all tasks
+SELECT task.title As Task, status.name AS Progress
+FROM task 
+INNER JOIN status ON task.status_id = status.id;
+
+-- 9. Get the name of each status, along with a count of how many tasks have that status
+Select status.name as Status_name, count(task.id) AS Status_name_count
+From status
+LEFT JOIN task on status.id = task.status_id
```
josejfernandez: The column `status_id` in the table `task` cannot be null (this can be checked in the `CREATE TABLE` statement [here](https://github.com/HackYourFuture-CPH/databases/blob/main/lesson1/lesson-data.sql#L24)). This means that this `LEFT JOIN` is no different from an `INNER JOIN` or `JOIN` (which is the same as an `INNER JOIN`).

This doesn't block the review either, but I suggest that you use `LEFT` or `RIGHT` joins only when they make a difference from an `INNER`, i.e.: when the column you are joining on can be null.

josejfernandez (approved):
Hey Farzaneh, nicely done! I checked all the statements and they worked correctly. I left two comments with suggestions, but nothing major that blocks the review. Good job!

I have noticed that my review does not allow you to merge - I have requested access to fix this.

josejfernandez (approved):

josejfernandez (approved):

databases/week1/Homework.sql
FarzanehAhmadi: Thanks for the suggestion! I've updated all SQL keywords to uppercase for consistency.

databases/week1/Homework.sql
FarzanehAhmadi: Done!

FarzanehAhmadi:
Hi Jose! Thanks for the feedback, I really appreciate it! ✨
Regarding merging, as far as I know, we don’t merge our PRs when working in this repository.

josejfernandez:
Your latest changes look great Farzaneh! Also, the way you have structured the changes in two independent commits with descriptive and accurate messages is very helpful when collaborating with other developers in a professional environment. Very well done there too!

I'm still figuring out how to best conduct the reviews in HYF - if there is anything else I can do for you, let me know :)

PR #147, opened 2025-02-05
Databases week1/soheib homework

siderdk:
Easy peasy 
database squeezy
Thank you very much in advance


databases/week1/create-db-template.sql
```diff
@@ -0,0 +1,113 @@
...
+insert into task (id, title, description, created, updated, due_date, status_id, user_id) values (32, 'Put up the new lamp in the hallway', NULL, '2017-10-15 05:45:54', '2017-10-16 14:05:35', '2017-12-29 02:29:26', 3, 3);
+insert into task (id, title, description, created, updated, due_date, status_id, user_id) values (33, 'Hang up paintings in living room', NULL, '2017-09-10 05:36:11', '2017-10-09 17:40:42', null, 3, 4);
+insert into task (id, title, description, created, updated, due_date, status_id, user_id) values (34, 'Buy plane ticket to Auckland', 'Check prices online first!', '2017-09-05 09:07:22', '2017-10-15 09:36:06', '2017-12-07 11:10:05', 1, 9);
+insert into task (id, title, description, created, updated, due_date, status_id, user_id) values (35, 'Learn about NoSQL databases', 'MongoDB, CouchDB, etc.', '2017-10-20 01:41:53', '2017-10-04 07:19:56', '2017-12-23 10:13:42', 2, NULL);
+
+SELECT COUNT(DISTINCT title)
```
josejfernandez, line 72: Interesting choice! The most common choice to count would be to count on the primary key of the table. I can see two things that we can discuss about this one:

**Correctness:** the exercise indicates _all_ tasks, not those who have different titles. By indicating `DISTINCT` you deliberately filter out those with the same title, which goes a bit against the requirement in the exercise of listing _all_ tasks :P so I would not add the `DISTINCT` constraint here.

**Performance:** I just checked the [query plan](https://dev.mysql.com/doc/refman/8.4/en/execution-plan-information.html) for both this query and the more common `SELECT COUNT(id) FROM task;`. It turns out, when counting on `DISTINCT title`, the database engine needs to read _all_ rows (which is called [full table scan](https://dev.mysql.com/doc/refman/8.4/en/glossary.html#glos_full_table_scan)) in order to determine how many with different titles are there. When counting on the primary key, the engine only needs to check the index, so it's much faster.

In this case it does not matter, but if you had many rows, your query would be noticeably slower than counting on the id. So, I would just count on the id :)

databases/week1/create-db-template.sql
```diff
@@ -0,0 +1,113 @@
...
+insert into task (id, title, description, created, updated, due_date, status_id, user_id) values (32, 'Put up the new lamp in the hallway', NULL, '2017-10-15 05:45:54', '2017-10-16 14:05:35', '2017-12-29 02:29:26', 3, 3);
+insert into task (id, title, description, created, updated, due_date, status_id, user_id) values (33, 'Hang up paintings in living room', NULL, '2017-09-10 05:36:11', '2017-10-09 17:40:42', null, 3, 4);
+insert into task (id, title, description, created, updated, due_date, status_id, user_id) values (34, 'Buy plane ticket to Auckland', 'Check prices online first!', '2017-09-05 09:07:22', '2017-10-15 09:36:06', '2017-12-07 11:10:05', 1, 9);
+insert into task (id, title, description, created, updated, due_date, status_id, user_id) values (35, 'Learn about NoSQL databases', 'MongoDB, CouchDB, etc.', '2017-10-20 01:41:53', '2017-10-04 07:19:56', '2017-12-23 10:13:42', 2, NULL);
+
+SELECT COUNT(DISTINCT title) 
+FROM `task`; /*all tasks 35*/
+
+SELECT COUNT(*)
```
josejfernandez, line 75: The exercises are worded like "Find all the tasks...", which could mean to count or to return all their information. Joining and filtering in the queries is perfect, so I don't think anything needs to be changed here - depends a bit on how the exercise is interpreted :)

josejfernandez (approved):
Hey Soheib, nicely done with the queries! I have noticed that you have put it a bit of effort to not do the "obvious" solutions. Kudos for this! I have dropped a few comments which I hope you would find informative. Keep up the good job, mate!

On the other hand, it has been a bit confusing to look at the pull request. There is work from previous weeks that I have not reviewed, and your homework is mixed in the same file with the database creation script.

I suggest that work from different tasks is separate in different branches, which is also a good practice for professional software developers.

My approval therefore goes _only_ for the SQL queries you have written, and does not extend to the remaining JavaScript code, which I haven't reviewed.

databases/week1/create-db-template.sql
siderdk: That is just really great advice and knowledge. Thank you so much. I will adjust

siderdk:
> Hey Soheib, nicely done with the queries! I have noticed that you have put it a bit of effort to not do the "obvious" solutions. Kudos for this! I have dropped a few comments which I hope you would find informative. Keep up the good job, mate!
> 
> On the other hand, it has been a bit confusing to look at the pull request. There is work from previous weeks that I have not reviewed, and your homework is mixed in the same file with the database creation script.
> 
> I suggest that work from different tasks is separate in different branches, which is also a good practice for professional software developers.
> 
> My approval therefore goes _only_ for the SQL queries you have written, and does not extend to the remaining JavaScript code, which I haven't reviewed.

Thank you so much for the feedback. 
The JS part was a mistake. I will clean it up. 
And I will implement the changes. thank you so much, this was very instructive

PR #148, opened 2025-02-05
database week1/jianxin

jianxinz233:
database week1 homework

### DIFF
--- a/databases/week1/database-hw-w1.sql
+++ b/databases/week1/database-hw-w1.sql
@@ -1,7 +1,3 @@
-SELECT * FROM users;
-SELECT * FROM task;
-SELECT * FROM status;
-
 -- Find out how many tasks are in the task table
 SELECT count(id) FROM task;
 
--- a/databases/week1/database-hw-w1.sql
+++ b/databases/week1/database-hw-w1.sql
@@ -7,11 +7,11 @@ WHERE due_date is NULL;
 
 -- Find all the tasks that are marked as done
 SELECT * FROM task 
-WHERE status_id IN (SELECT id FROM status WHERE name LIKE "%done%");
+WHERE status_id = (SELECT id FROM status WHERE name = 'Done');
 
 -- Find all the tasks that are not marked as done
 SELECT * FROM task
-WHERE status_id NOT IN (SELECT id FROM status WHERE name LIKE "%done%");
+WHERE status_id NOT IN (SELECT id FROM status WHERE name = 'Done');
 
 -- Get all the tasks, sorted with the most recently created first
 SELECT * FROM task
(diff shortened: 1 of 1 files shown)


databases/week1/database-hw-w1.sql
```diff
@@ -0,0 +1,46 @@
+-- Find out how many tasks are in the task table
+SELECT count(id) FROM task;
+
+-- Find out how many tasks in the task table do not have a valid due date
+SELECT count(id) FROM task
+WHERE due_date is NULL;
+
+-- Find all the tasks that are marked as done
+SELECT * FROM task 
+WHERE status_id IN (SELECT id FROM status WHERE name LIKE "%done%");
```
josejfernandez: This does not block the review because your query works, but I still suggest you follow this advice to get more clear and meaningful code.

You use `IN` when there is a list of possible values. Although it works in this case, there is only one status for "Done", so you could do `SELECT * FROM task WHERE status_id = (...)` and this statement would be a bit more clear.

Similarly, in the subquery you are searching for statuses whose name contain `done`. There is only one status for "Done", so you could do `SELECT id FROM status WHERe name = 'Done'`, which again is a bit more clear.

databases/week1/database-hw-w1.sql
```diff
@@ -0,0 +1,46 @@
...
+
+-- Get the title and due date of all tasks where the title or description contains database
+SELECT title, due_date FROM task
+WHERE title LIKE "%database%"
+OR description LIKE "%database%";
+
+-- Get the title and status (as text) of all tasks
+SELECT task.title AS task_title, status.name AS task_status
+FROM task
+LEFT JOIN status on task.status_id = status.id;
```
josejfernandez: The column `status_id` in the table `task` cannot be null (this can be checked in the `CREATE TABLE` statement [here](https://github.com/HackYourFuture-CPH/databases/blob/main/lesson1/lesson-data.sql#L24)). This means that this `LEFT JOIN` is no different from an `INNER JOIN` or `JOIN` (which is the same as an `INNER JOIN`).

This doesn't block the review either, but I suggest that you use `LEFT` or `RIGHT` joins only when they make a difference from an `INNER`, i.e.: when the column you are joining on can be null.

josejfernandez (approved):
Hey Jianxin, very well done! All your queries worked fine for me. I have written two comments with suggestions, that don't block the review. I still suggest that you address them, but it's up to you :)

jianxinz233:
> Hey Jianxin, very well done! All your queries worked fine for me. I have written two comments with suggestions, that don't block the review. I still suggest that you address them, but it's up to you :)

Thank you Jose for your suggestions! And I have updated my homework :)

PR #149, opened 2025-02-06
Databases week1/parisa

SeyedehParisaMousaviamiri:


databases/week1/database.sql
```diff
@@ -0,0 +1,44 @@
+SELECT COUNT(*) AS total_tasks FROM task;
```
kiko-bkr, line 3: Good job on using an alias with the `AS` keryword. 😉

databases/week1/database.sql
```diff
@@ -0,0 +1,44 @@
...
+FROM task 
+WHERE due_date IS NULL 
+
+SELECT * 
+FROM task 
+WHERE status_id = (SELECT id FROM status WHERE name = 'Done');
+
+SELECT * 
+FROM task 
+WHERE status_id != (SELECT id FROM status WHERE name = 'Done');
```
kiko-bkr, line 15: Nice work on using a [subquery](https://www.geeksforgeeks.org/mysql-subquery/)!

databases/week1/database.sql
```diff
@@ -0,0 +1,44 @@
...
+OR description LIKE '%database%';
+
+SELECT task.title, status.name AS status
+FROM task
+JOIN status ON task.status_id = status.id;
+
+SELECT status.name, COUNT(task.id) AS task_count
+FROM status
+LEFT JOIN task ON status.id = task.status_id
+GROUP BY status.name;
```
kiko-bkr, line 38: Amazing! You had the attention to detail to use `LEFT JOIN` here, instead of a simple `JOIN`. This is the correct way to do it, since this means that if a status does not have a task that would be associated with it, the status would still show up in the result with a `task_count` of 0. With a simple `JOIN` this would not happen. Well done!!! 🤩

databases/week1/database.sql
```diff
@@ -0,0 +1,44 @@
...
+
+SELECT status.name, COUNT(task.id) AS task_count
+FROM status
+JOIN task ON status.id = task.status_id
+GROUP BY status.name
+ORDER BY task_count DESC;
```
kiko-bkr, line 44: Same applies here, use a `LEFT JOIN` here as well.

kiko-bkr (approved):
Great work!! I just left a small suggestion. Keep it up! ⚡

databases/week1/database.sql
SeyedehParisaMousaviamiri: Thanks a lot, I appreciate your time and suggestions.

PR #150, opened 2025-02-06
Homework 1 is done

andrpavl:
DB connected finally) and homework is done

### DIFF
--- a/javascript/javascript1/week1/homework_1.sql
+++ b/javascript/javascript1/week1/homework_1.sql
@@ -1,0 +1,62 @@
+--1. Find out how many tasks are in the task table
+
+SELECT COUNT (*)
+FROM task
+
+--2. Find out how many tasks in the task table do not have a valid due date
+
+SELECT COUNT (*)
+FROM task 
+WHERE due_date IS NULL
+
+--3. Find all the tasks that are marked as done
+
+SELECT task.title, status.name
+FROM task JOIN status ON task.status_id = status.id
+WHERE status.name IN ('Done')
+
+--4. Find all the tasks that are not marked as done
+
+SELECT task.title, status.name
+FROM task JOIN status ON task.status_id = status.id
+WHERE status.name NOT IN ('Done')
+
+--5. Get all the tasks, sorted with the most recently created first
+
+SELECT * 
+FROM task
+ORDER BY created
+
+--6. Get the single most recently created task
+
+SELECT * 
+FROM task
+ORDER BY created
+LIMIT 1 
+
+--7. Get the title and due date of all tasks where the title or description contains database
+
+SELECT task.title, task.due_date
+FROM task 
+WHERE task.title LIKE '%database%' 
+   OR task.description LIKE '%database%';
+
+--8. Get the title and status (as text) of all tasks
+
+SELECT task.title, status.name
+FROM task JOIN status ON task.status_id = status.id
+
+--9. Get the name of each status, along with a count of how many tasks have that status
+
+SELECT status.name, COUNT(task.id)
+FROM status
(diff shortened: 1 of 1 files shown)


javascript/javascript1/week1/homework_1.sql
```diff
@@ -0,0 +1,62 @@
...
+
+SELECT COUNT (*)
+FROM task 
+WHERE due_date IS NULL
+
+--3. Find all the tasks that are marked as done
+
+SELECT task.title, status.name
+FROM task JOIN status ON task.status_id = status.id
+WHERE status.name IN ('Done')
```
kiko-bkr: As you are comparing with a single value, there is no need to check for the `IN` operator. It is enough (and more efficient) to just check for equality.

javascript/javascript1/week1/homework_1.sql
```diff
@@ -0,0 +1,62 @@
...
+
+--4. Find all the tasks that are not marked as done
+
+SELECT task.title, status.name
+FROM task JOIN status ON task.status_id = status.id
+WHERE status.name NOT IN ('Done')
```
kiko-bkr: Same applies here.

javascript/javascript1/week1/homework_1.sql
```diff
@@ -0,0 +1,62 @@
...
+
+--5. Get all the tasks, sorted with the most recently created first
+
+SELECT * 
+FROM task
+ORDER BY created
```
kiko-bkr: By default, MySQL orders elements in **ascending** order, not descending. This means that this query sorts with the smalles dates (i.e. earliest, least recent dates) first.

javascript/javascript1/week1/homework_1.sql
```diff
@@ -0,0 +1,62 @@
...
+
+--6. Get the single most recently created task
+
+SELECT * 
+FROM task
+ORDER BY created
+LIMIT 1
```
kiko-bkr, line 35: Same here, this returns the oldest task.

javascript/javascript1/week1/homework_1.sql
```diff
@@ -0,0 +1,62 @@
...
+
+SELECT task.title, task.due_date
+FROM task 
+WHERE task.title LIKE '%database%' 
+   OR task.description LIKE '%database%';
+
+--8. Get the title and status (as text) of all tasks
+
+SELECT task.title, status.name
+FROM task JOIN status ON task.status_id = status.id
```
kiko-bkr, line 47: It would have been nice to give the `status.name` field an alias with `AS` keyword. This way it shows up in the result as `name`, even though it represents a status.

javascript/javascript1/week1/homework_1.sql
```diff
@@ -0,0 +1,62 @@
...
+
+--9. Get the name of each status, along with a count of how many tasks have that status
+
+SELECT status.name, COUNT(task.id)
+FROM status
+LEFT JOIN task ON status.id = task.status_id
+GROUP BY status.name;
```
kiko-bkr, line 54: No need to make this `JOIN` left.

kiko-bkr, line 54: Amazing attention to detail with using `LEFT JOIN` instead of `JOIN` here! This means that if a status would not have a single task matching, it would still show up in the result with a zero value. Using a simple `JOIN` would mean that a status with a count of zero would not show in the result. Great work!!! 🎖️

kiko-bkr (changes requested):
There are some small changes needed, but great logic behind the queries, especially the last two! ⚡

kiko-bkr (approved):
Nice!

kiko-bkr (approved):

PR #152, opened 2025-02-07
databases-week1/guzide

GuzideGuzelbey:
Sorry for the late submission. 
Creating a connection to the database server on VSCode took so much time :)
But finally done. 

Thanks in advance!

### DIFF
--- a/databases/week1/week1-hw1.sql
+++ b/databases/week1/week1-hw1.sql
@@ -26,7 +26,7 @@ WHERE status_id in (
 );
 
 -- Get all the tasks, sorted with the most recently created first
-SELECT id, title, created 
+SELECT *
 FROM task
 ORDER BY created DESC;
 
--- a/databases/week1/week1-hw1.sql
+++ b/databases/week1/week1-hw1.sql
@@ -1,0 +1,62 @@
+-- Find out how many tasks are in the task table: 35
+SELECT COUNT(*) 
+FROM task;
+
+-- Find out how many tasks in the task table do not have a valid due date: 8
+SELECT COUNT(*) 
+FROM task 
+WHERE due_date IS NULL;
+
+-- Find all the tasks that are marked as done:12 tasks
+SELECT *
+FROM task
+WHERE status_id in (
+    SELECT id
+    FROM status
+    WHERE name = "done"
+);
+
+-- Find all the tasks that are not marked as done: 23
+SELECT *
+FROM task
+WHERE status_id in (
+    SELECT id
+    FROM status
+    WHERE name != "done"
+);
+
+-- Get all the tasks, sorted with the most recently created first
+SELECT id, title, created 
+FROM task
+ORDER BY created DESC;
+
+-- Get the single most recently created task: task with id=25
+SELECT id, title, created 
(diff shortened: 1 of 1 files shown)


databases/week1/week1-hw1.sql
```diff
@@ -0,0 +1,62 @@
...
+WHERE status_id in (
+    SELECT id
+    FROM status
+    WHERE name != "done"
+);
+
+-- Get all the tasks, sorted with the most recently created first
+SELECT id, title, created 
+FROM task
+ORDER BY created DESC;
```
kiko-bkr, line 31: No need to filter the fields here, just use `SELECT *`.

databases/week1/week1-hw1.sql
```diff
@@ -0,0 +1,62 @@
...
+SELECT title, due_date
+FROM task
+WHERE 
+title LIKE "%database%" OR 
+description LIKE "%database%"
+
+-- Get the title and status (as text) of all tasks
+SELECT task.title, status.name as status_name
+FROM task
+JOIN status ON task.status_id = status.id;
```
kiko-bkr, line 49: Good job on using an alias with `as status_name` here!

databases/week1/week1-hw1.sql
```diff
@@ -0,0 +1,62 @@
...
+
+-- Get the name of each status, along with a count of how many tasks have that status
+SELECT status.name as status_name, COUNT(task.id) as task_count
+FROM status
+LEFT JOIN task ON task.status_id = status.id
+GROUP BY status.name;
```
kiko-bkr, line 55: Great work on using `LEFT JOIN` here! That is perfect, since a normal `JOIN` would mean that any status that has no tasks would not show up in the result, even though it should! In your solution, using `LEFT JOIN`, it would show up, with a count of zero. Perfect! ❗

databases/week1/week1-hw1.sql
```diff
@@ -0,0 +1,62 @@
...
+
+-- Get the names of all statuses, sorted by the status with most tasks first
+SELECT status.name as statusname, COUNT(task.id) as taskcount
+FROM status
+LEFT JOIN  task on task.status_id = status.id
+GROUP BY status.name
+ORDER BY taskcount DESC;
```
kiko-bkr, line 62: Same here!

kiko-bkr (approved):
Amazing work. Only a small comment as a suggestion, but other than that your work is on point! 🏆

databases/week1/week1-hw1.sql
GuzideGuzelbey: thank you, done.

databases/week1/week1-hw1.sql
GuzideGuzelbey: Thank you learned from our mentor during the session and loved the idea :)

databases/week1/week1-hw1.sql
GuzideGuzelbey: Thank you :)

GuzideGuzelbey:
Thanks for your time :)

PR #154, opened 2025-02-08
homework- changed

Aminata00:
thanks in advance for the review !!

### DIFF
--- a/databases/week1/queries.sql
+++ b/databases/week1/queries.sql
@@ -1,0 +1,100 @@
+
+-- Insert Users
+INSERT INTO `users` (id, name, email, phone) VALUES
+(1, 'Aarika Ellingworth', 'aellingworth0@harvard.edu', '483-396-8795'),
+(2, 'Pren Goldsworthy', 'pgoldsworthy1@spotify.com', '635-572-8467'),
+(3, 'Pablo Kisbee', 'pkisbee2@lulu.com', '790-962-8683'),
+(4, 'Rodie Duncan', 'rduncan3@quantcast.com', '646-743-6191'),
+(5, 'Aubry Polak', 'apolak4@indiatimes.com', '302-678-7931'),
+(6, 'Maryrose Meadows', 'mmeadows5@comcast.net', '251-524-6594'),
+(7, 'Pavel Brushneen', 'pbrushneen6@techcrunch.com', '316-170-3640'),
+(8, 'Hedy Gerault', 'hgerault7@nymag.com', '176-177-5579'),
+(9, '王秀英', 'wang.xiuying@weebly.com', '891-952-6749'),
+(10, 'إلياس', 'elias@github.com', '202-517-6983'),
+(11, 'Donald Duck', 'donald@duck.com', NULL),
+(12, 'Adam Smith', 'smith@bla.com', NULL);
+
+-- Insert Statuses
+INSERT INTO `statuses` (id, name) VALUES
+(1, 'Not started'),
+(2, 'In progress'),
+(3, 'Done');
+
+-- Insert Tasks
+INSERT INTO `tasks` (id, title, description, created, updated, due_date, status_id, user_id) VALUES
+(1, 'Wash clothes', 'Title says it all.', '2017-10-25 06:54:16', '2017-10-15 13:05:09', NULL, 2, 1),
+(2, 'Become a billionaire', 'Invent a time machine and buy bitcoin in 2010.', '2017-09-26 03:06:46', '2017-10-08 06:14:31', '2017-12-22 20:58:03', 3, 6),
+(3, 'Plan meeting with London office', 'We will probably use Skype', '2017-10-04 18:07:37', '2017-10-14 16:01:31', '2017-12-05 19:42:15', 2, 8),
+(4, 'Order groceries online', 'The fridge is almost empty, we need eggs and milk', '2017-09-20 19:34:43', '2017-10-15 23:35:45', '2017-12-24 16:00:46', 1, 1),
+(5, 'Empty the mailbox', NULL, '2017-09-27 15:17:08', '2017-10-08 17:31:16', NULL, 2, 9),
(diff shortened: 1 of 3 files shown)


databases/week1/queries.sql
```diff
@@ -0,0 +1,100 @@
...
+(11, 'Donald Duck', 'donald@duck.com', NULL),
+(12, 'Adam Smith', 'smith@bla.com', NULL);
+
+-- Insert Statuses
+INSERT INTO `statuses` (id, name) VALUES
+(1, 'Not started'),
+(2, 'In progress'),
+(3, 'Done');
+
+-- Insert Tasks
```
kiko-bkr, line 23: You could have used the data from the lesson: [Link](https://github.com/HackYourFuture-CPH/databases/blob/main/lesson1/lesson-data.sql)
That way one of the exercises would have returned a result, since right now it does not.

databases/week1/queries.sql
```diff
@@ -0,0 +1,100 @@
...
+SHOW TABLES;
+
+-- Find out how many tasks are in the task table
+SELECT COUNT(*)
+FROM tasks;
+
+-- Find out how many tasks in the task table do not have a valid due date
+SELECT * 
+FROM tasks 
+WHERE due_date= "NULL";
```
kiko-bkr, line 50: This is not the right way to check for rows where a field is NULL. You have to use something else than `=`, and no need for quotation marks around `NULL`.

databases/week1/queries.sql
```diff
@@ -0,0 +1,100 @@
...
+
+
+-- Find all the tasks that are marked as done
+SELECT title 
+FROM tasks 
+WHERE status_id = 3;
```
kiko-bkr, line 56: No need to filter just for `title`.

databases/week1/queries.sql
```diff
@@ -0,0 +1,100 @@
...
+
+-- Find all the tasks that are not marked as done
+ELECT *
```
kiko-bkr:
```suggestion
SELECT * 
```

databases/week1/queries.sql
```diff
@@ -0,0 +1,100 @@
...
+
+-- Find all the tasks that are marked as done
+SELECT title 
+FROM tasks 
+WHERE status_id = 3;
+
+-- Find all the tasks that are not marked as done
+ELECT * 
+FROM tasks 
+WHERE status_id = 3;
```
kiko-bkr, line 61: This finds the tasks that ARE done.

databases/week1/queries.sql
```diff
@@ -0,0 +1,100 @@
...
+
+-- Get all the tasks, sorted with the most recently created first
+SELECT title, due_date
+FROM tasks 
+ORDER BY DESC
```
kiko-bkr, line 66: No need to filter for `title` and `due_date`.
You do not `ORDER BY` an actual field, thus the query has an error.

databases/week1/queries.sql
```diff
@@ -0,0 +1,100 @@
...
+
+
+-- Get the single most recently created task
+SELECT title
+FROM task 
+WHERE due_date =  MAX(due_date)
```
kiko-bkr, line 72: This would be the more optimal way, but you need a [subquery](https://www.geeksforgeeks.org/mysql-subquery/) to find the max value first.

databases/week1/queries.sql
```diff
@@ -0,0 +1,100 @@
...
+-- or i also thought , but idk if its optimal 
+SELECT * FROM tasks
+ORDER BY due_date DESC
+LIMIT 1;
```
kiko-bkr, line 76: This works well. 👍

databases/week1/queries.sql
```diff
@@ -0,0 +1,100 @@
...
+
+
+-- Get the title and due date of all tasks where the title or description contains database
+SELECT title, due_date 
+WHERE title LIKE '%database%'
+OR description LIKE '%database%';
```
kiko-bkr, line 82: You did not specify which table to query `FROM`. Also, this is the query that returns no rows as you did not use the lesson data.

databases/week1/queries.sql
```diff
@@ -0,0 +1,100 @@
...
+
+-- Get the title and status (as text) of all tasks
+SELECT title, statuses.name AS stat_text
+FROM tasks
+JOIN statuses ON tasks.status_id = statuses.id;
```
kiko-bkr, line 87: 👏

databases/week1/queries.sql
```diff
@@ -0,0 +1,100 @@
...
+
+-- Get the name of each status, along with a count of how many tasks have that status
+SELECT statuses.name AS stat_name, COUNT(tasks.id) AS task_count
+FROM statuses
+JOIN tasks ON statuses.id = tasks.status_id
+GROUP BY stat_name;
```
kiko-bkr, line 93: 👏

databases/week1/queries.sql
```diff
@@ -0,0 +1,100 @@
...
+
+-- Get the names of all statuses, sorted by the status with most tasks first
+SELECT statuses.name AS stat_name, COUNT(tasks.id) AS task_count
+FROM statuses
+JOIN tasks ON statuses.id = tasks.status_id
+GROUP BY stat_name
+ORDER BY task_count DESC;
```
kiko-bkr, line 100: 👏

kiko-bkr (changes requested):
Needs some fixes, but the queries are overall well written!

databases/week1/queries.sql
Aminata00: hi @kiko-bkr !! I had some errors with the syntax on that file so I used the updated version sent on slack instead;))

databases/week1/queries.sql
Aminata00: will check it out!

Aminata00:
> Needs some fixes, but the queries are overall well written!

thanks for the great review!! will make the changes asapppp

PR #155, opened 2025-02-08
Javascript javascript3 week3/jalil

jalilhu:
Finalized Quiz Project


databases/week1/homeWork.sql
```diff
@@ -0,0 +1,92 @@
+create DATABASE hyf_db;
```
te-online, line 1: This file does not belong here, please move it to the appropriate branch ♻️ 😊

javascript/javascript2/week3/index.html
```diff
@@ -0,0 +1,34 @@
...
+  <head>
+    <meta charset="UTF-8" />
+    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+    <title>Quiz App</title>
+  </head>
+
+  <body>
+    <p>Redirecting to <a href="quizzForm.html">Create 4 choice Quiz</a>...</p>
+
+    <main>
```
te-online, line 12: I'm not sure why this `main` element is even here, if you're only redirecting to `quizzForm.hml` on this page?

javascript/javascript2/week3/index.html
```diff
@@ -0,0 +1,34 @@
...
+          <div class="player">
+            <h3>Multiplayer</h3>
+            <img src="#" alt="multiPlayer" />
+            <button id="to-multiplayer" class="to-game">Go</button>
+          </div>
+        </div>
+      </section>
+    </main>
+
+    <script type="module" src="./scripts/script.js">
```
te-online, line 30: You don't need to specify `src` when you're just executing a redirect in your script 😊

javascript/javascript2/week3/scripts/multiPlayer.js
```diff
@@ -0,0 +1,149 @@
...
+
+    questionTitle[1].textContent = question.question;
+    quantity[1].textContent = `${currentQuestionIndex + 1} out of ${
+      questions.length
+    }`;
+    pointHolder[1].textContent = `Correct: ${correctPoints}`;
+    questionNumber[1].textContent = `Incorrect: ${incorrectPoints}`;
+    answer = question.answer;
+
+    let value = await new Promise((resolve) => {
```
te-online, line 86: I'm curious why you're using a Promise here? I can see how this could work, but it's a bit unconventional and can lead to _very_ hard to find bugs. Instead, you could try making a function that you can pass as callback to `addEventListener`. This function can then execute the code that now is executed after the promise resolves. This also allows you to remove the event listener to avoid memory leaks.

javascript/javascript3/week3/scripts/quizToSaveApi.js
```diff
@@ -0,0 +1,73 @@
+
+
+const owner = "jalilhu";
+
+const repo = "Quiz_App";
+const path = "userQuestions.json"; // Path to the JSON file
+const token = ""
```
te-online, line 7: I like the idea of being able to save updates to your questions! That being said, this will unfortunately not work. JavaScript files are visible to the browser and the user using that browser. You access token for Github is privat and needs to remain private. Hence, you can never publish a website that directly talks to Github from the JS file that way. However, you can use a server to accomplish that goal. Then the server will know the secret and check if a user has access to the specific action and forward it to the Github API.

te-online (commented):
A couple of files in, I realised that this PR includes `javascript2` files. I skipped them from then on. You also included an SQL file. Please move files to the correct branches before opening a PR; always check the "Files" tab in Github to see which files will be included in the PR. You will get a much better review if you're making life easier for the reviewer 💟 

Let's move on to my feedback for you! 👇 

Generally, I'd like to approve this PR, but it should _not_ be merged to this repository. You were supposed to create your own repository and you can still do that and copy over your code and commit it there. Using your own repository, you can more easily showcase it to others.

- 🎮  I really like the aesthetics of your quiz app; you put a lot of effort into splashy effects and making it feel like an actual game. Well done!
- 💅  Your files could use formatting and comments, to make it easier for others to read your code and for you to understand your code in the future.
- ⚠️ You have a lot of seemingly duplicate files and duplicate code between files. Please clean this up and move the result into a personal repository. 
- ✅ You seem to already have deployed parts of your app to Github pages (your JSON), which is nice. So you already seem to have a repository. Please move the code there as well and deploy your app to showcase it as a project in the future.

I'm going to stop my review here. I did not go through all of the files; as mentioned there seems to be a lot of repetition and I don't really see an advantage for you to get feedback on files you may not even use.

I've tried you app locally and it looks very good, I like the functionality 👍

PR #156, opened 2025-02-08
Databases week1/jalil

jalilhu:
sql queries


databases/week1/homeWork.sql
```diff
@@ -0,0 +1,92 @@
...
+
+SELECT * FROM tasks WHERE due_date IS NULL
+
+ALTER TABLE tasks ADD is_done BOOLEAN;
+
+UPDATE tasks
+SET is_done = CASE 
+    WHEN status = "DONE" THEN TRUE  
+    ELSE FALSE
+END;
```
kiko-bkr, line 41: Instead of using this `UPDATE` here, you should have used this exact condition (`WHERE status = 'DONE') in the `SELECT` queries. Then there is no need to UPDATE the table.

databases/week1/homeWork.sql
```diff
@@ -0,0 +1,92 @@
...
+
+SELECT title, is_done FROM hyf_db.tasks WHERE is_done = TRUE ;
+
+SELECT title, is_done FROM hyf_db.tasks WHERE is_done = FALSE ;
+
+
+SELECT MIN(created_at) AS earliest_created_at, title, description 
+FROM tasks 
+GROUP BY title, description 
+ORDER BY earliest_created_at ASC;
```
kiko-bkr, line 59: No need for these commands.

databases/week1/homeWork.sql
```diff
@@ -0,0 +1,92 @@
...
+
+SELECT created_at, title, description from tasks ORDER BY created_at ASC
+LIMIT 1
```
kiko-bkr, line 62: Good job!

databases/week1/homeWork.sql
```diff
@@ -0,0 +1,92 @@
...
+
+UPDATE tasks
+SET due_date = '2025-02-28'
+WHERE id in (11)
+
+
+SELECT title, due_date from tasks where title like '%database%' OR description like '%database%' GROUP BY title, due_date
```
kiko-bkr, line 69: No need for the `GROUP BY`, a great query otherwise! ✅

databases/week1/homeWork.sql
```diff
@@ -0,0 +1,92 @@
+create DATABASE hyf_db;
+use hyf_db;
+
+CREATE TABLE tasks (
```
kiko-bkr, line 4: You should use the database schema (meaning the tables) and data from the [lesson](https://github.com/HackYourFuture-CPH/databases/tree/main/lesson1). An important part of the homework is to practice joining tables, and the way you defined the `tasks` table, there is no need for `JOIN` commands.

databases/week1/homeWork.sql
```diff
@@ -0,0 +1,92 @@
...
+
+SELECT
+    CASE 
+        WHEN is_done = true THEN 'DONE'  
+        ELSE  'NOT_DONE'
+    END AS status_name,
+    COUNT(*) AS task_count
+FROM tasks
+GROUP BY is_done
+ORDER BY task_count DESC
```
kiko-bkr, line 89: These queries do not use `JOIN` commands, as the status is already in the `tasks` table based on your definition. Rewrite them using the lesson plan data that I linked in the other comment.

kiko-bkr (changes requested):
Great initiative! I see that you are practicing different aspects of SQL already. The main issue is that you defined a different `tasks` table than the one used in class. If you rewrite the queries with that data, you will be on the right track! 💪

PR #158, opened 2025-02-10
Add files via upload

StefanyRomeroV:


databases/week1/Homework1
```diff
@@ -0,0 +1,51 @@
...
+WHERE due_date is NULL;
+
+-- Find all the tasks that are marked as done
+
+SELECT * FROM task
+WHERE status_id = (SELECT id FROM status WHERE name = 'Done');
+
+--Find all the tasks that are not marked as done
+SELECT * FROM task
+WHERE status_id NOT IN (select id FROM status WHERE name ='Done')
```
josejfernandez, line 17: This is a small thing that does not block the review, given that the query works correctly. Normally, `IN` and `NOT IN` are used with a list of values. In this case, you expect the subquery to return only one item and you could therefore use other operator that communicates the intent of the query a bit better:

```
SELECT * FROM task WHERE status_id != (...)
```

databases/week1/Homework1
```diff
@@ -0,0 +1,51 @@
...
+--Get the title and status (as text) of all tasks
+SELECT task.title AS task_title, status.name AS task_status
+FROM task
+INNER JOIN status on task.status_id = status.id;
+
+--Get the name of each status, along with a count of how many tasks have that status
+SELECT status.name  AS task_status,
+count(task.title) AS task_count
+FROM task
+INNER JOIN status ON task.status_id=status_id
```
josejfernandez, line 42: This query does not work as intended! Please give it another go :) Pay attention to the columns you are joining on: what tables do they come from? In which table is the `status_id` column?

databases/week1/Homework1
```diff
@@ -0,0 +1,51 @@
...
+GROUP BY task_status;
+
+--Get the names of all statuses, sorted by the status with most tasks first
+SELECT status.name AS task_status,
+count(task.title) AS task_count
+FROM task
+iNNER JOIN status ON  task.status_id = status_id
```
josejfernandez, line 49: Same as previous query: this one does not work as intended.

josejfernandez (changes requested):
Hey Stefany! All of your queries worked perfectly, except the last two. Please fix them so they don't shadow your otherwise great job with the queries 😀

PR #159, opened 2025-02-10
Database week1/yuusuf

Ganja0003:


databases/week1/data.sql
```diff
@@ -0,0 +1,62 @@
+-- 1. Find out how many tasks are in the task table
+SELECT count(*) 
+from task;
+
+
+-- 2. Find out how many tasks in the task table do not have a valid due date
+SELECT count (*)
```
josejfernandez, line 8: Beware, this returns an error when I try to run it:

```
[42000][1064] You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near '*)
```

Can you spot the cause? Look at the first query, that one works.

databases/week1/data.sql
```diff
@@ -0,0 +1,62 @@
...
+-- 2. Find out how many tasks in the task table do not have a valid due date
+SELECT count (*) 
+from task 
+WHERE due_date is NULL;
+
+
+-- 3. Find all the tasks that are marked as done
+SELECT * 
+FROM task 
+WHERE status_id = 3; -- 3 is equal to done
```
josejfernandez, line 16: Smart shortcut! No need to change this, because it works fine, just a comment: you could have also joined tables (as you did in query 8) and then filtered on the status name:

```
SELECT *
FROM `task`
    INNER JOIN `status` ON `task`.`status_id` = `status`.`id`
WHERE `status`.`name` = 'Done';
```

josejfernandez (approved):
Hey Yuusuf! Good job with the queries 💪🏻 With the exception of the second query, all the others worked perfectly fine. Well done!

PR #161, opened 2025-02-11
Databases week2/farzaneh

FarzanehAhmadi:


databases/week2/homework2.sql
```diff
@@ -0,0 +1,129 @@
+
+--Part 1:
+-- Add a task with these attributes: title, description, created, updated, due_date, status_id, user_id
+INSERT INTO task (title, description, created, updated, due_date, status_id, user_id)
+VALUES
+('Do sql homework', 'Check the github!','2025-2-10 17:54:16','2025-2-25 06:54:16','2025-3-13 06:54:16', 2, NULL);
```
josejfernandez, line 6: There is no need to change anything here, because I tested the query and it worked fine, so just a comment. The date format normally uses two digits for the month: `2025-02-10`. While it worked with MySQL if you omit the leading zero, maybe other databases won't be so tolerant and require that you type the two digits for the month.

Keep it in mind while handling dates in databases (and code!).

databases/week2/homework2.sql
```diff
@@ -0,0 +1,129 @@
...
+
+-- Delete a task
+DELETE FROM task
+WHERE id = 5;
+
+
+-- Part2: School database
+---class table
+CREATE TABLE `class`(
+  `id` int(10) unsigned NOT NULL AUTO_INCREMENT PRIMARY KEY,
```
josejfernandez, line 36: I think this is a nice choice of type for the primary key 💯

databases/week2/homework2.sql
```diff
@@ -0,0 +1,129 @@
...
+);
+
+---Student table
+CREATE TABLE `student`(
+  `id` int(10) unsigned NOT NULL AUTO_INCREMENT PRIMARY KEY,
+  `name` varchar(255) NOT NULL,
+  `email` varchar(255) NOT NULL,
+  `phone` varchar(255) NULL,
+  `class_id` INT(10) UNSIGNED NOT NULL,
+    CONSTRAINT `fk_class` FOREIGN KEY (`class_id`) REFERENCES `class` (`id`) ON DELETE CASCADE ON UPDATE CASCADE
```
josejfernandez, line 49: Well done with the constraint! In real world outside of homework, it is advisable to be careful with `ON DELETE CASCADE` behaviors, though, since they can mercilessly delete a lot of stuff from the database 😃

databases/week2/homework2.sql
```diff
@@ -0,0 +1,129 @@
...
+);
+
+---Create an index on the name column of the student table.
+CREATE INDEX index_name
+ON student (name);
+
+---Add a new column to the class table named status which can only have the following values: not-started, ongoing, finished
+ALTER Table class
+ADD COLUMN status ENUM('not-started','ongoing','finished') NULL;
```
josejfernandez, line 58: Careful, you are setting the field to accept `NULL`, which from my perspective deviates from the requirement of this exercise - I think this should not accept `NULL` values.

databases/week2/homework2.sql
```diff
@@ -0,0 +1,129 @@
...
+JOIN `user` on `user`.id = user_task.user_id
+JOIN status on task.status_id = status.id
+WHERE `user`.name = 'Donald Duck' AND status.name = 'Not started';
+
+-- Get all the tasks for 'Maryrose Meadows' that were created in september (hint: month(created)=month_number)
+SELECT task.title , task.created, `user`.name
+FROM task
+JOIN user_task ON task.id = user_task.task_id
+JOIN `user` on `user`.id = user_task.user_id
+WHERE `user`.name = 'Maryrose Meadows' AND MONTH(task.created) = 10;
```
josejfernandez: Wow Farzaneh, the year is already long enough with 12 months, let's not add more 😅 double check what number would be correct for September.

databases/week2/homework2.sql
```diff
@@ -0,0 +1,129 @@
...
+    event_id INT AUTO_INCREMENT PRIMARY KEY,
+    name VARCHAR(255) NOT NULL,
+    event_date DATETIME NOT NULL,
+    venue_id INT NOT NULL,
+    organizer_id INT NOT NULL,
+    CONSTRAINT `fk_venues` FOREIGN KEY (venue_id) REFERENCES venues(venue_id),
+    CONSTRAINT `fk_organizer` FOREIGN KEY (organizer_id) REFERENCES organizers(organizer_id)
+);
+
+CREATE TABLE guests (
```
josejfernandez, line 120: Food for thought, in case you want to learn a bit more: how could you control in this database that no excess guests are added to an event, based on the venue capacity?

In other words, if certain event is associated with a venue with a capacity of 100, how could we tell MySQL to not allow new rows to be inserted on the `guests` table for that event, once there are 100 guests?

A solution could be to use _triggers_ `BEFORE INSERT`, as can be read [here](https://www.geeksforgeeks.org/mysql-before-insert-trigger/).

josejfernandez (approved):
Everything looks good to me Farzaneh, you are kicking ass in Hack Your Future! 💯 There are some things that could be improved slightly, but overall it's mainly minor things. Good effort!

databases/week2/homework2.sql
```diff
@@ -0,0 +1,129 @@
...
+--Part 1:
+-- Add a task with these attributes: title, description, created, updated, due_date, status_id, user_id
+INSERT INTO task (title, description, created, updated, due_date, status_id, user_id)
+VALUES
+('Do sql homework', 'Check the github!','2025-2-10 17:54:16','2025-2-25 06:54:16','2025-3-13 06:54:16', 2, NULL);
+
+-- Change the title of a task
+UPDATE task
+SET title = 'Update Data for Homework Assignment'
+WHERE task.title = 'Make the databases perform better'
```
josejfernandez: It's fine to update based on the title, but keep in mind that more than one task could have the same title, given the current structure of the table.

databases/week2/homework2.sql
FarzanehAhmadi: Good point! I’ll look into triggers, thanks for the tip!

databases/week2/homework2.sql
FarzanehAhmadi: Haha, fixed it! 😅

databases/week2/homework2.sql
FarzanehAhmadi: Done!

FarzanehAhmadi:
Hi Jose! Thanks for the feedback, I really appreciate it! ✨

PR #163, opened 2025-02-12
Database week1/ayman

Aymanngu:
Sorry for the late upload. I had major issues trying to find out how to connect/setup SQL, and I finally figured it out.
At least I hope that I have done it right :D


josejfernandez (approved):
Hey Ayman! You absolutely nailed it! All works flawlessly and the code is beautifully formatted too, 💯 ! Excellent job!

PR #165, opened 2025-02-12
Databases week2/guzide

GuzideGuzelbey:
All tasks completed, thank you in advance :)

UPDATE: These 2 commits are not valid anymore, since I changed them after the feedbacks I received in the last session of the database module.([a63c795], [50892c6])


databases/week2/hw-part1.sql
```diff
@@ -0,0 +1,35 @@
...
+SET due_date=NULL
+WHERE id=32;
+
+-- fourth task
+
+UPDATE task
+SET status_id=3
+WHERE id=1;
+
+--fifth task / not sure if I understood the task correctly
```
NVedsted, line 23: I can see why the wording could be a bit confusing. The task intended for you to set the status of a task to the existing status "Done". This is very close  though with the added benefit of practicing another insert! 😄

databases/week2/hw-part2.sql
```diff
@@ -0,0 +1,28 @@
...
+    `begins` DATETIME NOT NULL,
+    `ends` DATETIME NOT NULL
+);
+
+CREATE TABLE student (
+    `id` int(10) unsigned NOT NULL AUTO_INCREMENT PRIMARY KEY,
+    `name` varchar(255) NOT NULL,
+    `email` varchar(255) NOT NULL,
+    `phone` varchar(255) NULL,
+    `class-id` int(10) unsigned NOT NULL,
```
NVedsted, line 18: Good job with the new tables!

A tiny nitpick is that I'd suggest sticking to underscores (`_`) as `-` could be confused with subtraction and might even be an illegal identifier in some programming languages! 😄

databases/week2/hw-part3.sql
```diff
@@ -0,0 +1,171 @@
...
+
+SELECT t.*
+FROM task t
+JOIN user_task ut ON t.id = ut.task_id 
+JOIN user u ON ut.user_id = u.id
+WHERE u.name = "Maryrose Meadows"
+AND MONTH(t.created) = 9;
+
+SELECT
+    MONTH(created) AS month_number,
```
NVedsted, line 167: Nice queries and good use of aliasing to make the results easier to read! ⭐

databases/week2/hw-part4.sql
```diff
@@ -0,0 +1,71 @@
+CREATE DATABASE mycompany_db;
+USE mycompany_db;
+
+SET NAMES utf8mb4;
+
+CREATE TABLE `emp` (
```
NVedsted, line 6: An interesting data model to work with! I would be careful with shortened names like `emp` and `dep`  as it might not be super clear to people what it means.

It is also not possible to run these CREATE statements as they are laid out since `emp` depends on `job` and `dep` which do not exist at this point. This also applies to your insertions where employees are inserted before their jobs and departments are. This can be solved by simply reordering the statements. 😄

NVedsted (approved):
Good job with the hand-in! It was very nicely structured hand-in. 🤩 I have added a few comments for you to take a look at.

databases/week2/hw-part4.sql
GuzideGuzelbey: Ah, I did not know that it was also executing from top to bottom, since I had "run" button for each snippet in vs code. But now it makes sense, of course :DDD

And thank you for the  warning with the table naming.

databases/week2/hw-part2.sql
GuzideGuzelbey: Thank you :-)

PR #168, opened 2025-02-13
Databases week2/jalil

jalilhu:
Thanks for reviewing my pull request
and thanks for leaving feedbacks

I used 2 VSCode extesions 

1- Draw.io Integration
2- vscode-pdf


databases/week2/more_queries.sql
```diff
@@ -0,0 +1,53 @@
...
+select * from tasks AS ta
+inner JOIN
+`user` as us
+where us.email like '%@spotify.com%'
+
+
+SELECT t.*, u.email 
+FROM task t
+LEFT JOIN `user` u ON t.user_id = u.id
+WHERE u.email LIKE '%@spotify.com%';
```
MaherHussain, line 12: good using left join, so i think you don't need first one with inner join

MaherHussain:
good job. 
the best practice is to use lowercase for naming the tables

PR #171, opened 2025-02-15
Databeses-week2/hossein

hossein1065:


databases/week2/week2.sql
```diff
@@ -0,0 +1,97 @@
...
+-- part 1
+SELECT * from status
+SELECT * FROM task;
+
+INSERT into task(title, description, created, updated, due_date, status_id, user_id)
+VALUES('Database week2','Do homework for database week 2', '2025-02-09 12:00:00', '2025-02-13 17:10:15','2025-02-12 18:00:00',3,1);
+
+UPDATE task
+SET title = "HUF homework"
+WHERE title= "Database week2";
```
NVedsted, line 11: Nice and simple statements; excellent! ⭐

As a small note, I'd be wary of updating on anything but a unique identifier as you might update more data than you expect or less for that matter. I can see you're switching to IDs below, so I am probably preaching to the choir. 😄

databases/week2/week2.sql
```diff
@@ -0,0 +1,97 @@
...
+
+
+DELETE from task
+WHERE id=16;
+
+-- part 2
+
+CREATE DATABASE sql_School;
+USE sql_School;
+CREATE Table Classes;
```
NVedsted, line 34: The intent behind these CREATE statements seems clear and correct, but sadly I could not run them due to syntax errors and it not being able to find a table called `classes`. Can you see why and what you need to change? 😄

databases/week2/week2.sql
```diff
@@ -0,0 +1,97 @@
...
+
+
+
+ALTER Table classes 
+ADD status  ENUM ('not-started', 'ongoing', 'finished');
+
+--part 3
+use hyf_lesson2;
+
+select *
```
NVedsted, line 63: If you were to add another user with a Spotify-email and add them to one of Pren's tasks, this query would show the same task twice. Can you tell why and how to avoid that? 😄 

Also this task is asking for tasks, so it'd be easier to read the result if it only shows task-related columns, i.e., limiting the `*` selector a bit.

databases/week2/school.sql
```diff
@@ -0,0 +1,61 @@
+CREATE DATABASE IF NOT EXISTS School;
```
NVedsted, line 1: Interesting data model which iseasy to follow! 👍 

The intent of the script also seems to be correct, but you cannot run the script as-is on an empty database. Can you tell why?

A small nitpick is that the column `instructors_id` on `courses` can be a bit misleading as it is only a single instructor.

NVedsted (changes requested):
Nice hand-in! ⭐ In particularly, I enjoyed the structure as it was easy to see what task was being solved. 😄 

There are some issues though which I have highlighted in the comments. 🙂

PR #173, opened 2025-02-15
Databases week2/soheib

siderdk:
Dear Mentor, 
Below are the homework committed files. 
For the part 4, I used figjam, but when it came to the sql part I ended up introducing few modifications. I will adjust the ERD at a later stage. 
Thank you in advance for your very appreciated time, energy, and feedback.


databases/week2/homework2.sql
```diff
@@ -0,0 +1,23 @@
+USE my_db;
+
+-- Part1: working with tasks
+INSERT INTO task (title, description, created, updated, due_date, status_id)
+VALUES 
+('home work', 'do the home work', NOW(), NOW(), '2020-12-13', 2);
+
+INSERT INTO user_task (user_id, task_id)
+SELECT 10, id FROM task WHERE title = 'home work' LIMIT 1;
```
kiko-bkr, line 9: Going the extra mile to also add an association to the new task huh? 😉  Nicely done, going beyond the exercise.

databases/week2/homework2.sql
```diff
@@ -0,0 +1,23 @@
...
+
+UPDATE task
+SET title = 'Home work task updated'
+WHERE title = 'home work';
+
+UPDATE task
+SET due_date = '2020-12-12', status_id = 3, description = 'completed'
+WHERE title = 'Home work task updated';
```
kiko-bkr, line 17: The `description` does not have to be changed here, only the status.

databases/week2/homework2_part2.sql
```diff
@@ -0,0 +1,22 @@
...
+    email VARCHAR(255) NOT NULL,
+    phone VARCHAR(255) NOT NULL,
+    class_id INT,
+    FOREIGN KEY (class_id) REFERENCES class(id) ON DELETE CASCADE
+);
+
+CREATE INDEX index_name ON student(name);
+
+ALTER TABLE class 
+ADD COLUMN status ENUM('not-started', 'ongoing', 'finished') NOT NULL DEFAULT 'not-started';
```
kiko-bkr, line 22: Good job on providing a default value. 👏

databases/week2/homework2_part3.sql
```diff
@@ -0,0 +1,19 @@
...
+SELECT * FROM task
+JOIN user_task ON user_task.task_id = task.id
+JOIN user ON user.id = user_task.user_id
+JOIN status ON status.id = task.status_id
+WHERE user.name = 'Donald Duck' AND status.name = 'Not started';
+
+SELECT * FROM task
+JOIN user_task ON user_task.task_id = task.id
+JOIN user ON user.id = user_task.user_id
+WHERE user.name = 'Maryrose Meadows' AND MONTH(task.created) = 09;
```
kiko-bkr, line 15: It is fine to use the value `9` for the month as well. This is only a nit, though.

databases/week2/homework2_part3.sql
kiko-bkr, line 1: Great queries.

databases/week2/homework2_part4.sql
```diff
@@ -0,0 +1,52 @@
+CREATE TABLE users (
+    id VARCHAR(255) PRIMARY KEY,
```
kiko-bkr, line 2: It is advised against to use a string as a primary key, especially when the value of it is not critical to the logic of the schema. The necessary information about the entities in your schema are in the fields already, so there is no need to make their primary keys 255 length strings. 

If your idea was that such a primary key is guaranteed to be unique, or wanted to have a [UUID](https://en.wikipedia.org/wiki/Universally_unique_identifier), this can be done in a separate column, and the primary key can stay as a simple integer, as it is way more efficient to use an integer for indexing and searching.

databases/week2/homework2_part4.sql
```diff
@@ -0,0 +1,52 @@
...
+
+CREATE TABLE shopping_list (
+    id VARCHAR(255) PRIMARY KEY,
+    user_id VARCHAR(255) NOT NULL,
+    recipe_id VARCHAR(255) NOT NULL,
+    ingredient_id VARCHAR(255) NOT NULL,
+    measurement_unit_id VARCHAR(255) NOT NULL,
+    ingredient_amount DECIMAL(10, 2) NOT NULL,
+    FOREIGN KEY (user_id) REFERENCES users(id) ON DELETE CASCADE,
+    FOREIGN KEY (recipe_id) REFERENCES recipes(id) ON DELETE CASCADE,
```
kiko-bkr, line 49: I suppose the idea here is that each user would have a separate shopping list for each recipe. Interesting! 😮

databases/week2/Tweat-ERD.png
kiko-bkr, line 1: Why are the `recipes` and `users` tables connected here?

kiko-bkr (changes requested):
Great job Soheib! ⚡ 

I have left some small comments about your code, minor details.

When it comes to the schema of your designed database, I have read your comment on the pull request, and I assume the question of mine about the design image is what you were referring to. Moreover, I advise against using `VARCHAR()` for primary keys.

Address the comments. ✋

databases/week2/homework2_part4.sql
siderdk: Aha! the UUID was exactly my thought. 
so if I understand correctly, I keep the id an INT but then add a column for UUID where I can give each element a UUID. 
so that means that the id will still identify the element in the said table. but the UUID will be referring to the said element in any other context. 
Do I get it correctly?
I also heard of a UUID node package. I assume I can use for this.

databases/week2/homework2_part4.sql
siderdk: Yes, in this way when you get your shopped items, you can then store them in an effective way to help you prep faster for your recipes :P 
I have been promising my wife to build this app :D

databases/week2/Tweat-ERD.png
siderdk: Thank you for pointing out the mistake. 
I have also few unnecessarily duplicated fields. 
Now that I remember we addressed this in class the following week

PR #174, opened 2025-02-15
Database-week2/yuusuf

Ganja0003:
The diagram was really challenging, but that's good. The part that confused me was the different relationships between the entities. All in all, it was a good and beneficial exercise to understand the relationships between them.


databases/week2/task3.sql
```diff
@@ -0,0 +1,28 @@
+--Get all the tasks assigned to users whose email ends in @spotify.com
+select * from task
+join user_task on task_id = user_task.task_id
+join user on user_task.task_id = user.id
+where user.email like '%@spotify.com';
+
```
MaherHussain, line 6:
```suggestion

SELECT t.*, u.email 
FROM task t
LEFT JOIN `user` u ON t.user_id = u.id
WHERE u.email LIKE '%@spotify.com%';
```

PR #179, opened 2025-02-19
Database -week3/marjan

najmeh25:
add database homewok


databases/week3/homework3.sql
```diff
@@ -0,0 +1,170 @@
...
+--Get meals that have a price smaller than a specific price (e.g., 90)
+SELECT title,price 
+FROM meals 
+WHERE price < 90;
+
+--Get meals that still have available reservations
+SELECT m.id, m.title, m.description, m.location, m.price
+FROM meals m
+LEFT JOIN reservations r ON m.id = r.meal_id
+GROUP BY m.id ;
```
frej0887, line 131: In this task, you only join the two tables together, you don't actually use `reservations` for anything, so you don't actually solve this task.

databases/week3/homework3.sql
```diff
@@ -0,0 +1,170 @@
+-- Active: 1739038387593@@127.0.0.1@3306@meal_sharing
+--Create the Database and Tables
```
frej0887, line 2: Great with a description of the tasks, but it is not quite a valid comment, since you need a space after the ´--´ in MySql

databases/week3/homework3.sql
```diff
@@ -0,0 +1,170 @@
...
+
+--Get all meals
+SELECT * FROM meals;
+
+--Add a new meal
+INSERT INTO meals(title,description,location,price,max_reservations)
+VALUES('Pizza Night', 'Delicious homemade pizza', 'Copenhagen', 75.00, 10);
+
+--Get a meal with any id
+SELECT * FROM meals WHERE id=5;
```
frej0887, line 55: It would be nice if you could write your assignment in such a way, so it is runable from top to bottom without any changes. In this case, you can't select a meal with id=5 when you only have one meal in your database

databases/week3/homework3.sql
```diff
@@ -0,0 +1,170 @@
+-- Active: 1739038387593@@127.0.0.1@3306@meal_sharing
+--Create the Database and Tables
+CREATE database meal_sharing;
+use meal_sharing;
+CREATE TABLE users (
+    id INT AUTO_INCREMENT PRIMARY KEY,
+    name VARCHAR(255) NOT NULL,
+    email VARCHAR(255) UNIQUE NOT NULL
+);
```
frej0887, line 9: You've decided to add a table for users instead of just using an email like the assignment asked for. 
It is fun to add an extra thing, but you don't use it as well as you could've. You could for example when querying for reservations, make it include the name and email

```sql
SELECT r.meal_id, r.created_date, r.number_of_guests, u.name, u.email
FROM reservations r
JOIN users u ON u.id = r.user_id;
```

Of course you didn't need to add this table, so this is just a suggestion :)

frej0887 (changes requested):
Hi! Generally nice assignment, but I think you missed something in one of the last tasks.
I really like your creativity though

PR #180, opened 2025-02-19
add homework week3

ruslanaonshchk:
add part meal

### DIFF
--- a/databases/week3/additional_querries.sql
+++ b/databases/week3/additional_querries.sql
@@ -1,0 +1,25 @@
+select * from Meals where price < 90;
+
+select Meals.* from Meals
+LEFT JOIN Reservations ON Meals.id = Reservations.meal_id
+GROUP BY Meals.id
+HAVING COUNT(Reservations.id) < Meals.max_reservations;
+
+select * from Meals where title like '%Rød grød med%';
+
+select * from Meals where created_date BETWEEN '2023-01-01' AND '2023-12-31';
+
+select * from Meals limit 5;
+
+select Meals.* from Meals
+INNER JOIN Reviews ON Meals.id = Reviews.meal_id
+where Reviews.stars >= 4
+GROUP BY Meals.id;
+
+select * from Reservations where meal_id = 1 ORDER BY created_date;
+
+select Meals.*, AVG(Reviews.stars) AS average_rating
+from Meals
+LEFT JOIN Reviews ON Meals.id = Reviews.meal_id
+GROUP BY Meals.id
+ORDER BY average_rating DESC;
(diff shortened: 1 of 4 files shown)


databases/week3/reviews.sql
```diff
@@ -0,0 +1,23 @@
...
+  stars INT NOT NULL,
+  created_date TIMESTAMP DEFAULT CURRENT_TIMESTAMP,
+  FOREIGN KEY (meal_id) REFERENCES Meals(id) ON DELETE CASCADE
+);
+
+select * from Reviews;
+
+INSERT INTO Reviews (meal_id, name, review, stars) 
+select id, 'Alice', 'Amazing meal and great company!', 5 
+from Meals where id = 1;
```
frej0887, line 15: This works, but you can make the query simpler like this

```sql
INSERT INTO Reviews (meal_id, name, review, stars)
VALUES (3, 'Alice', 'Amazing meal and great company!', 5)
```

Had the setup been more complex, this could make sense, but not in this case. You are asking it to pull out the id from the Meals table, where the id is 1, so asking for the id while also giving it, which is double the work

databases/week3/additional_querries.sql
```diff
@@ -0,0 +1,25 @@
+select * from Meals where price < 90;
+
+select Meals.* from Meals
+LEFT JOIN Reservations ON Meals.id = Reservations.meal_id
+GROUP BY Meals.id
+HAVING COUNT(Reservations.id) < Meals.max_reservations;
```
frej0887, line 6: This code works, but you are mixing an aggregated value with one which isn't, which  you are not supposed to do. You can either put `Meals.max_reservations` into the group by clause or the nicer solution, make `COUNT(Reservations.id)` not aggregated. A way to do that is to change your join into another query, so `where Meals.max_reservations > (select count(*) from Meals where Meals.id=Reservation.meal_id)`.  

It is a bit hard to explain, so text me in Slack if you want a better explanation.

databases/week3/meal.sql
```diff
@@ -0,0 +1,23 @@
...
+  max_reservations INT NOT NULL,
+  price DECIMAL(10, 2) NOT NULL,
+  created_date TIMESTAMP DEFAULT CURRENT_TIMESTAMP
+);
+
+select * from Meals;
+
+INSERT INTO Meals (title, description, location, max_reservations, price) 
+select 'Pasta Night', 'Enjoy a homemade Italian dinner', 'Copenhagen', 10, 12.50 
+from DUAL;
```
frej0887, line 15: You just taught me something new, I didn't know of this table. 
My last comment explains how to do it simpler

frej0887 (changes requested):
Nice job with the handin!
It would be nice if you could run the code from top to bottom, always getting some values. So always make sure to add an item before trying to read it. 
It would also be nice if you added a comment with the task above the code, so it is easier for me to read :)
```
-- Get all meals                  <-- This is the comment, starting with `-- `
select * from Meals;
```

PR #183, opened 2025-02-20
Database-week3/hossein

hossein1065:


databases/week3/reserve_meal.sql
```diff
@@ -2,3 +2,147 @@ CREATE DATABASE meal_reserve_share;
...
+from meal
+LIMIT 5;
+
+SELECT m.title, m.price,r.stars
+FROM meal m
+JOIN review  r
+on m.id= r.meal_id
+where r.stars=(SELECT MAX(stars) 
+    FROM review
+    WHERE meal_id = m.id);
```
frej0887, line 135: I think you might've misunderstood this task, you are supposed to find meals with good reviews, but what you found here is the meals with the best review rating. I think you were supposed to find good reviews, of lets say 4 stars or more. With the way you've made it, if there are only bad reviews of 1-2 stars, then your query would return meals with 2 stars, which are the best ones, but not exactly good. 
But it is a nice solution to the task you assumed it was

databases/week3/reserve_meal.sql
```diff
@@ -2,3 +2,147 @@ CREATE DATABASE meal_reserve_share;
...
+
+
+select * 
+from reservation
+where id=2
+ORDER by create_date;
```
frej0887, line 141: This is supposed to be reservations for a specific meal, but you've accidentally sorted on reservation id, not meal id. Since reservation id is unique, this query will always return either 0 (if the id is not in use) or 1 row.

databases/week3/reserve_meal.sql
```diff
@@ -2,3 +2,147 @@ CREATE DATABASE meal_reserve_share;
...
+INSERT INTO meal(title,description,location,meal_time,max_reservation,price,create_date)
+VALUES('Lasagna',' is a type of pasta','italian','2025-02-22 08:20:00',30,70,'2025-02-22');
+
+
+SELECT * 
+from meal
+WHERE id=1;
+
+UPDATE meal
+SET title= "kobideh kabab", price=80, max_reservation=60
```
frej0887, line 55: Both ' and " works for marking text as strings, but the convention is to use '

frej0887 (changes requested):
Well done one the homework!
It would be nice if your code is a bit more consistent with capitalization of table names and keywords and whitespace, or had the task names as a comment above the task
`-- Get meals that still has available reservations`
`code...`

hossein1065:
Hi. Thanks for the feedback. I don’t know why I didn’t pay attention to
that part ☺️. I will try to fix it. Thanks for your time

PR #184, opened 2025-02-20
Database week3/yuusuf

Ganja0003:
Done with this week's homework and looking forward to the next module.


databases/week3/additional_queries.sql
```diff
@@ -0,0 +1,48 @@
...
+
+
+
+-- Get only specific number of meals fx return only 5 meals
+SELECT * from meal limit 2;
+
+
+
+-- Get the meals that have good reviews
+SELECT meal.title AS meal,meal.description,meal.location,review.title as review,review.description,review.stars from meal
```
frej0887, line 30: This would be a lot easier to read if you would put each select onto a new line like this:
```sql
SELECT
    meal.title AS meal,
    meal.description,
    meal.location,
    review.title as review,
    review.description,
    review.stars 
from meal
```

databases/week3/db.sql
```diff
@@ -0,0 +1,74 @@
...
+    FOREIGN KEY (meal_id) REFERENCES meal(id)
+) 
+
+
+
+
+
+--data insertion for meal table
+insert into meal (title,description,location,`when`,max_reservations,price,created_date)
+VALUES ('Lasagna', 'Enjoy a delicious lasagna', 'Rome', '2025-05-01 10:00:00', 10, 50, '2025-02-15');
```
frej0887, line 46: I really like that you are adding all these nice examples from the start

databases/week3/additional_queries.sql
```diff
@@ -0,0 +1,48 @@
+-- Get meals that has a price smaller than a specific price fx 90
+SELECT * from meal where price < 70;
+
+
+
+-- Get meals that still has available reservations
+SELECT meal.* FROM meal
+LEFT JOIN reservation ON reservation.meal_id = meal.id
+GROUP BY meal.id
+HAVING COUNT(reservation.id) < meal.max_reservations;
```
frej0887, line 10: This code works, but you are mixing an aggregated value with one which isn't, which you are not supposed to do. You can either put `Meals.max_reservations` into the group by clause or the nicer solution, make `COUNT(Reservations.id)` not aggregated. A way to do that is to change your join into another query, so `where Meals.max_reservations > (select count(*) from Meals where Meals.id=Reservation.meal_id)`.

It is a bit hard to explain, so text me in Slack if you want a better explanation.

frej0887 (commented):
I had a few comments, but nothing major, nice work!

PR #185, opened 2025-02-20
databases-week3/jianxin

jianxinz233:
hand in homework week 3

### DIFF
--- a/databases/week3/meal_sharing_app.sql
+++ b/databases/week3/meal_sharing_app.sql
@@ -1,0 +1,177 @@
+USE meal_sharing_app;
+
+CREATE TABLE `meal` (
+    `id` INT(10) UNSIGNED NOT NULL AUTO_INCREMENT PRIMARY KEY,
+    `title` VARCHAR(255) NOT NULL,
+    `description` TEXT DEFAULT NULL,
+    `location` VARCHAR(255) NOT NULL,
+    `when` DATETIME NOT NULL,
+    `max_reservations` INT(10) UNSIGNED NOT NULL,
+    `price` DECIMAL(10,2) NOT NULL,
+    `created_date` DATETIME NOT NULL
+);
+
+CREATE TABLE `reservation` (
+    `id` INT(10) UNSIGNED NOT NULL AUTO_INCREMENT PRIMARY KEY,
+    `number_of_guests` INT(10) UNSIGNED NOT NULL,
+    `meal_id` INT(10) UNSIGNED NOT NULL,
+    `created_date` DATETIME NOT NULL,
+    `contact_phonenumber` VARCHAR(255) NOT NULL,
+    `contact_name` VARCHAR(255) NOT NULL,
+    `contact_email` VARCHAR(255) NOT NULL,
+    CONSTRAINT `fk_meal_reservation` FOREIGN KEY (`meal_id`) REFERENCES `meal` (`id`) ON DELETE CASCADE
+);
+
+CREATE TABLE `review` (
+    `id` INT(10) UNSIGNED NOT NULL AUTO_INCREMENT PRIMARY KEY,
+    `title` VARCHAR(255) NOT NULL,
+    `description` TEXT DEFAULT NULL,
+    `meal_id` INT(10) UNSIGNED NOT NULL,
+    `stars` INT(10) UNSIGNED NOT NULL,
+    `created_date` DATETIME NOT NULL,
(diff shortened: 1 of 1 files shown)


databases/week3/meal_sharing_app.sql
```diff
@@ -0,0 +1,177 @@
...
+INSERT INTO meal 
+    (`title`, `description`, `location`, `when`, `max_reservations`, `price`, `created_date`)
+VALUES 
+    ('Pizza', 'Traditional italien pizza', 'Aarhus', '2025-3-1', 6, 100, '2025-2-13'),
+    ('Asian Food', 'Fusion asian experience', 'Aarhus', '2025-3-2', 6, 96, '2025-2-14'),
+    ('Open Sandwich', 'Fresh Danish indregients', 'Aarhus', '2025-3-3', 6, 120, '2025-2-15'),
+    ('Salad Bowl', 'A healthy way of delicious bites', 'Aarhus', '2025-2-25', 6, 89, '2025-2-16'),
+    ('Big Burger', 'American Style Burger', 'Aarhus', '2025-2-28', 4, 160, '2025-2-17');
+
+--Get a meal with any id, fx 1
```
frej0887, line 50: You need a space after `--` to make a comment, which you forgot to add here

databases/week3/meal_sharing_app.sql
```diff
@@ -0,0 +1,177 @@
...
+
+-- Add a new review
+INSERT INTO review
+    (title, description, meal_id, stars, created_date)
+VALUES
+    ('Great experience', 'Nice food and staffs, hygge place', 3, 5, '2025-3-6'),
+    ('Highly recommended', 'Great place for gathering with nice food', 2, 5, '2025-3-2'),
+    ('Recommended', 'Friendly to family with kids', 1, 5, '2025-3-3'),
+    ('Nice burger', 'Great burgurs, nice staffs', 5, 4, '2025-3-7'),
+    ('Eat healthily', 'Fresh ingredients', 4, 5, '2025-3-2');
```
frej0887, line 105: You write your dates with only a single digit for month or day. This works, but the code would look nicer if you 0-pad the month and day, so it says '2025-03-02', this way it has the same length as '2025-10-25'. 

If you really want it to look nice, you can order the way you insert into your fields, such that the known-width elements are first, which makes it easier to read. 
```
INSERT INTO review
    (created_date, stars, meal_id, title, description)
VALUES
    ('2025-03-06', 5, 3, 'Great experience', 'Nice food and staffs, hygge place'),
    ('2025-03-02', 5, 2, 'Highly recommended', 'Great place for gathering with nice food'),
    ('2025-10-17', 4, 5, 'Nice burger', 'Great burgurs, nice staffs'),
    ('2025-03-02', 5, 4, 'Eat healthily', 'Fresh ingredients');
```
All of this is completely optional though, and the code as is, is already nice

databases/week3/meal_sharing_app.sql
```diff
@@ -0,0 +1,177 @@
...
+
+-- Reservation
+-- Get all reservations
+SELECT * FROM reservation;
+
+-- Add a new reservation
+INSERT INTO reservation 
+    (number_of_guests, meal_id, created_date, contact_phonenumber, contact_name, contact_email)
+VALUES 
+    (2, 1, '2025-2-23', '12345678', 'Alice', 'alice@example.com'),
```
frej0887, line 73: This has meal id 1, which you just deleted above on line 62. You don't need to be able to run it in the order of the tasks, but it would be nice

databases/week3/meal_sharing_app.sql
```diff
@@ -0,0 +1,177 @@
...
+    `price` DECIMAL(10,2) NOT NULL,
+    `created_date` DATETIME NOT NULL
+);
+
+CREATE TABLE `reservation` (
+    `id` INT(10) UNSIGNED NOT NULL AUTO_INCREMENT PRIMARY KEY,
+    `number_of_guests` INT(10) UNSIGNED NOT NULL,
+    `meal_id` INT(10) UNSIGNED NOT NULL,
+    `created_date` DATETIME NOT NULL,
+    `contact_phonenumber` VARCHAR(255) NOT NULL,
```
frej0887, line 19: You don't need 255 chars to store a phone number. Limiting to the amount of digits in a phone number would be too much though, because you in the future might want to extend your business with international customers, and then you need to be able to store area codes or similar. You could safely cut it down to something like 20 chars

frej0887 (approved):
Very nice submission! I made a few comments but they are generally fairly unimportant and just helps you make your code a bit nicer✨

jianxinz233:
Thank you for your review and such good suggestions! I will fixed them and keep in mind for the future code. ❤️

PR #189, opened 2025-02-21
Databases week3/guzide

GuzideGuzelbey:
Thank you in advance :)

### DIFF
--- a/databases/week3/mealsharing-data.sql
+++ b/databases/week3/mealsharing-data.sql
@@ -64,7 +64,7 @@ UPDATE CASCADE
 
 insert into Meal
   (title, description, location,
-  `when`, max_reservations, price, created_date) 
+`when`, max_reservations, price, created_date) 
 VALUES
 ('Zeytinyağli Sarma', 'Rice stuffed wine leaves with olive oil, lemon juice and spices', 'Istanbul', '2025-02-27 15:30:00', 5, 59.50, '2024-01-01 10:38:09'),
 ('Dalyan Köfte', 'In the shape of a substantial meatloaf with boiled eggs in it, garnished with peas and carrots in tomato sauce', 'Ankara', '2025-06-25 12:30:00', 2, 189.00, '2024-01-01 10:40:00'),
@@ -105,27 +105,27 @@ FROM `Meal
 -- Add a new meal
 insert into Meal
   (title, description, location,
-  `when`, max_reservations, price, created_date) 
+`when`, max_reservations, price, created_date) 
 VALUE
 ('Ispanakli Börek', 'Flaky and creamy, this pastry combines crisp phyllo dough with tangy Turkish white cheese and tender spinach', 'Tekirdağ', '2025-04-24 12:30:00', 6, 125.00, '2023-09-01 18:08:09');
 -- Get a meal with any id, fx 1
 SELECT *
-FROM meal
+FROM Meal
 WHERE id=3;
 
 -- Update a meal with any id, fx 1. Update any attribute fx the title or multiple attributes
-UPDATE meal
+UPDATE Meal
 SET location = 'Tosya', price = 210
 WHERE id = 2;
 
 -- Delete a meal with any id, fx 1
-DELETE FROM meal 
+DELETE FROM Meal 
 WHERE id = 3;
 
 --Reservation Tasks
 -- Get all reservations
 SELECT *
-FROM reservation;
+FROM Reservation;
 
 -- Add a new reservation
 insert into Reservation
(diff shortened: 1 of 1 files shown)


databases/week3/mealsharing-data.sql
```diff
@@ -0,0 +1,200 @@
...
+  /* Get meals that partially match a title. 
+  Rød grød med will match the meal with the title Rød grød med fløde */
+SELECT *
+FROM meal
+WHERE title LIKE '%sü%';
+
+  -- Get meals that has been created between two dates
+SELECT id, title, created_date
+FROM meal
+WHERE '2023-08-10 10:30:25' < created_date <'2024-11-10 00:30:25';
```
frej0887: Your intent is good, unfortunately, the MySQL syntax does not support this, and you should instead split it into two parts with one `<` each.

This is what MySQL reads, step by step:

- `'2023-08-10 10:30:25' < created_date <'2024-11-10 00:30:25'`
- `('2023-08-10 10:30:25' < created_date) <'2024-11-10 00:30:25'` (order of operations in MySQL)
- `(0 or 1) <'2024-11-10 00:30:25'` (Evaluate comparison)
- `(0 or 1) < 2024` (MySQL converts the right side to a number.)
- `1` (Evaluate final comparison)

databases/week3/mealsharing-data.sql
```diff
@@ -0,0 +1,200 @@
...
+("Okayish", "Dalyan Köfte was okay, but I found it a bit too oily. Not sure I-d order it again.", 2, 2, "2025-02-03 11:26:14"),
+("Good but not great", "Dalyan Köfte was tasty, but I found it a little too salty. Still worth trying!", 2, 3, "2025-02-18 17:29:55"),
+("Not bad, but expected more", "Dalyan Köfte was decent, but I was hoping for stronger flavors. It was a bit bland.", 2, 4, "2025-02-14 19:51:09"),
+("Loved it!", "This was my first time trying Dalyan Köfte, and it won't be my last. Amazing balance of spices!", 2, 5, "2025-02-07 20:45:32"),
+("Num num", "I had never tried Kuru-Pilav before, but wow, what an explosion of flavors! Highly recommended.", 3, 5, "2025-02-12 09:34:45"),
+("Not as promoted", "I am dissapointed a little, unfort Kuru-Pilav didn't impress me. A bit too dry for me.", 3, 2, "2025-02-01 07:52:11"),
+("Super authentic!", "This tasted just like real home-cooked food. Kuru-Pilav was a surprisingly delicious blend of two different taste", 3, 5, "2025-02-17 16:30:50"),
+("Could be better", "The Kuru-Pilav was decent, but I felt like it was too spicy. Burned till my ears.", 3, 3, "2025-02-06 22:15:29");
+
+--Meal Tasks
```
frej0887: Comments are always nice, but this is not valid syntax, since MySQL expects a space after a `--` comment

databases/week3/mealsharing-data.sql
```diff
@@ -0,0 +1,200 @@
...
+
+  -- Add a new review
+insert into Review (title, description, meal_id, stars, created_date) 
+VALUES
+('I am burning!', 'I was not informed about the alergens,I have hives on my skin and my face is burning... Never Recommend!', 5, 1, '2025-02-02 09:30:12');
+
+  -- Get a review with any id, fx 1
+SELECT *
+FROM review
+WHERE id=8;
```
frej0887: In some places you write your tablenames with the first letter in capital case, other places you don't. The code would be nicer if it was more consistent. 
Be aware that MySQL run on a Windows pc does not care about the case of the titles of your tables, but Linux does.
See [this StackOverflow answer](https://stackoverflow.com/questions/2009005/are-column-and-table-name-case-sensitive-in-mysql) for more

frej0887 (changes requested):
Nice work on the handin, I especially like the layout, since I found it very readable

As a note, it would be nice if your code could be run from top to bottom.  Look at the foreign key in `Reservation`, which refers to `Meal` below it.

databases/week3/mealsharing-data.sql
GuzideGuzelbey: Thank you, updated.

databases/week3/mealsharing-data.sql
GuzideGuzelbey: Wow, amazing information. Thank you so much, updated 🤍

databases/week3/mealsharing-data.sql
GuzideGuzelbey: Thank you, updated.

GuzideGuzelbey:
Thank you for your time and insightful feedback  :raised_hands:

PR #192, opened 2025-02-24
Node week1/andrii p

andrpavl:


javascript/javascript1/week1/homework_1.sql
kiko-bkr, line 1: This file should not be in this directory, and also not in this pull request. :)

nodejs/week1/avg.js
```diff
@@ -0,0 +1,15 @@
+const inputValues = process.argv.slice(2);
+const numbersFromValues = inputValues.map((item) => Number(item));
```
kiko-bkr, line 2: Very nice to extract these values into variables with explicative names. Way better than using `process.argv.length` all the time. Great! 👏

nodejs/week1/avg.js
```diff
@@ -0,0 +1,15 @@
...
+
+if (numbersFromValues.some((num) => isNaN(num))) {
+	console.error("Please, enter valid numbers");
+	process.exit(1);
```
kiko-bkr, line 6: Very nice to use this function! This is exactly how we should let the user know that the code did not succeed.

nodejs/week1/avg.js
```diff
@@ -0,0 +1,15 @@
...
+}
+
+if (inputValues.length === 0) {
+	console.log("Enter some numbers for calculating");
+} else {
+	const sum = numbersFromValues.reduce((acc, item) => acc + item, 0);
+	const result = sum / numbersFromValues.length;
+	console.log(result);
```
kiko-bkr, line 14: Nice and advanced way to implement this! 🤩

kiko-bkr (approved):
Great job! No complaints at all! ⚡

PR #193, opened 2025-02-24
nodejs week1

jalilhu:
Hiiiiiii

### DIFF
--- a/nodejs/week1/avg.js
+++ b/nodejs/week1/avg.js
@@ -1,0 +1,13 @@
+function getAverage(a=0,b=0,c=0) {
+    const array = [a,b,c]
+    let sum = 0;
+    for (let i = 0; i < array.length; i++) {
+      sum += array[i];
+    }
+    const avg = Math.floor(sum/array.length)
+    console.log(`Average of a: ${a} and b: ${b} and c: ${c} = ${avg}`)
+    return Number(avg);
+  }
+  const args = process.argv.slice(2).map(Number); // Convert arguments to numbers
+
+  getAverage(...args);


nodejs/week1/avg.js
```diff
@@ -0,0 +1,13 @@
+function getAverage(a=0,b=0,c=0) {
```
kiko-bkr, line 1: This restricts the function to only be able to handle maximum three values. You can just use an array as the parameter of the function, and then it can handle any amount of values.

nodejs/week1/avg.js
```diff
@@ -0,0 +1,13 @@
...
+    const array = [a,b,c]
```
kiko-bkr, line 2: You can just have the `array` variable as the parameter of the function.

nodejs/week1/avg.js
```diff
@@ -0,0 +1,13 @@
...
+    let sum = 0;
+    for (let i = 0; i < array.length; i++) {
+      sum += array[i];
+    }
+    const avg = Math.floor(sum/array.length)
```
kiko-bkr, line 7: No need to floor this value, it can be a float as it is an average.

nodejs/week1/avg.js
```diff
@@ -0,0 +1,13 @@
...
+    console.log(`Average of a: ${a} and b: ${b} and c: ${c} = ${avg}`)
+    return Number(avg);
```
kiko-bkr, line 9: If you already print the result, no need to return it as well.

nodejs/week1/avg.js
```diff
@@ -0,0 +1,13 @@
+function getAverage(a=0,b=0,c=0) {
+    const array = [a,b,c]
+    let sum = 0;
+    for (let i = 0; i < array.length; i++) {
+      sum += array[i];
+    }
+    const avg = Math.floor(sum/array.length)
+    console.log(`Average of a: ${a} and b: ${b} and c: ${c} = ${avg}`)
```
kiko-bkr, line 8:
```suggestion
    console.log(`Average of the provided numbers is ${avg}`)
```

nodejs/week1/avg.js
```diff
@@ -0,0 +1,13 @@
...
+    return Number(avg);
+  }
+  const args = process.argv.slice(2).map(Number); // Convert arguments to numbers
+
+  getAverage(...args);
```
kiko-bkr, line 13:
```suggestion
  getAverage(args);
```
No need to use the `...` operator if the function accepts an array. (which it should)

kiko-bkr (changes requested):
Hi! I see that you had the impression that you are supposed to implement this in a way that it always expects 3 values, because of the example in the homework. However, you should implement this in a way that it can take any number of values.

I left some suggestions to get you started. 🙌 😉

PR #194, opened 2025-02-24
Nodejs week1/guzide

GuzideGuzelbey:
I somehow messed up with the git I guess, so if you can ignore the first 6 commit (database hw) it would be perfect.
Thank you in advance :)


nodejs/week1/package.json
kiko-bkr, line 1: As you did not have to create a server for this task, there is no need to install `express`, nor to create an NPM application. A single JavaScript file is enough. 😄

nodejs/week1/avg.js
```diff
@@ -0,0 +1,18 @@
...
+  let sum = nums.reduce(
+    (accumulator, currentValue) => accumulator + currentValue,
+    0
+  );
+  return sum / nums.length;
+}
+
+let args = process.argv.slice(2);
+
+let numbers = args.map(Number);
```
kiko-bkr, line 27: Good use of explanatory variables! They make the code a lot clearer, compared to cryptic variables like `process.argv`. 👏 

Note: Bear in mind that if a certain argument is not a number, mapping it with `Number` will result in `NaN`, and then the average will also be `NaN`. It was an extra question to handle this case, so no worries that you did not handle it, I am just letting you know. 😃

nodejs/week1/avg.js
```diff
@@ -0,0 +1,18 @@
+// https://medium.com/@ryan_forrester_/average-of-an-array-in-javascript-how-to-guide-85e97dd1443e
+// ref for the average calculation
+
+function calculateAverage(nums) {
+  if (nums.length === 0) return 0;
```
kiko-bkr: Good error handling. However, it would be nice to let the user know why he got zero as a result, and advise the user to actually provide the proper arguments.

nodejs/week1/avg.js
```diff
@@ -0,0 +1,18 @@
...
+  let sum = nums.reduce(
+    (accumulator, currentValue) => accumulator + currentValue,
+    0
+  );
```
kiko-bkr, line 21: Nice use of `reduce`!

kiko-bkr (changes requested):
Good job! I like the way you implemented the task, it is very easy to read and follow. 💯 

However, I do suggest that you let the user know if they are not using your code correctly, for example if they did not provide any arguments.

nodejs/week1/package.json
GuzideGuzelbey: Yes, may bad. I realized that after I completed this one and started the actual hw :D

nodejs/week1/avg.js
GuzideGuzelbey: Thank you, done.

nodejs/week1/avg.js
GuzideGuzelbey: Thank you :)

nodejs/week1/avg.js
GuzideGuzelbey: You are right, I haven't realized the extra part :]
Done.

GuzideGuzelbey:
Thank you for the feedback :)

kiko-bkr (approved):
Well done! ⚡

PR #196, opened 2025-02-25
Nodejs week1/farzaneh

FarzanehAhmadi:


nodejs/week1/avg.js
```diff
@@ -0,0 +1,18 @@
+const args = process.argv.slice(2);
+//console.log(typeof args);
+const numbers = args.map(arg => parseFloat(arg));
```
kiko-bkr, line 3: Nice idea to use some variables to make the code easier to understand. Variables like `process.argv` is hard to understand.

nodejs/week1/avg.js
```diff
@@ -0,0 +1,18 @@
...
+
+if(numbers.length === 0){
+  console.log("Please enter some numbers!");
+  process.exit();
+}
+if(numbers.some(isNaN)){
+  console.log("Error: All inputs must be valid numbers.")
+  process.exit();
+}
```
kiko-bkr: Great error handling. In order to let the user know that the program did not succeed, we should return an error code, for example `process.exit(1);`.

nodejs/week1/avg.js
```diff
@@ -0,0 +1,18 @@
...
+let sum = 0;
+numbers.forEach(i => {
+  sum += i;
+});
```
kiko-bkr, line 16: Good use of `forEach`! Another function to use here is `reduce()`. But both of them are great.

kiko-bkr (approved):
Well done! No complaints, outside of a small piece of advice to make your code even easier to use by the user, by returning an appropriate error code.

FarzanehAhmadi:
Thanks a lot for your feedback, I really appreciate it! ✨

PR #198, opened 2025-02-25
Create avg.js

Aminata00:
homework node warmup

### DIFF
--- a/nodejs/week1/avg.js
+++ b/nodejs/week1/avg.js
@@ -1,0 +1,27 @@
+/*/create a file named avg.js which will calculate the 
+average of the numbers provided as command line arguments./*/
+const input = process.argv.slice(2).map(Number); 
+
+export default function getAverage(...numbers) { 
+    let sum = 0;
+    for (let i = 0; i < numbers.length; i++) {
+        sum += numbers[i]; 
+    }
+    const average = sum / numbers.length;
+    return average;
+};
+
+console.log(getAverage(...input));
+
+// Optional improvements and considerations:
+//filter the values we pass thru getaverage 
+const numbers = input.filter(num => !isNaN(num)); 
+//put conditions? 
+if (input.length === 0) {
+    console.log("Please provide at least one number.");
+    process.exit(1); 
+}
+if (numbers.length === 0) {
+    console.log("No valid numbers provided.");
+    process.exit(1);
+}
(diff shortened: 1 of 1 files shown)


Aminata00:
Thanks in advance for the reviews !

nodejs/week1/avg.js
```diff
@@ -0,0 +1,27 @@
+/*/create a file named avg.js which will calculate the 
+average of the numbers provided as command line arguments./*/
+const input = process.argv.slice(2).map(Number); 
+
+export default function getAverage(...numbers) {
```
kiko-bkr, line 5: No need to export this, it is fine to just have it as a function. Also, as you are passing an array, you can also omit the `...` operator.
```javascript
function getAverage(numbers) {
```

nodejs/week1/avg.js
```diff
@@ -0,0 +1,27 @@
...
+export default function getAverage(...numbers) { 
+    let sum = 0;
+    for (let i = 0; i < numbers.length; i++) {
+        sum += numbers[i]; 
+    }
+    const average = sum / numbers.length;
+    return average;
+};
+
+console.log(getAverage(...input));
```
kiko-bkr:
```suggestion
console.log(getAverage(input));
```

nodejs/week1/avg.js
```diff
@@ -0,0 +1,27 @@
...
+
+// Optional improvements and considerations:
+//filter the values we pass thru getaverage 
+const numbers = input.filter(num => !isNaN(num));
```
kiko-bkr, line 18: Nice! That is a great way to avoid using values that are not numbers in our summing logic. 🧠  In this case, if the user passed values that are not numbers, they would just get ignored.

nodejs/week1/avg.js
```diff
@@ -0,0 +1,27 @@
...
+const numbers = input.filter(num => !isNaN(num)); 
+//put conditions? 
+if (input.length === 0) {
+    console.log("Please provide at least one number.");
+    process.exit(1); 
+}
+if (numbers.length === 0) {
+    console.log("No valid numbers provided.");
+    process.exit(1);
+}
```
kiko-bkr, line 27: Yes, this is also a great safeguard to handle the case where the user does not provide any parameters, or no valid numbers. This would result in a zero division error, so it's nice to protect against it!

kiko-bkr (approved):
Well done Aminata! I see that you first implemented the basic functionality, and then added the extra logic that would implement the extra features needed.

In the future feel free to just submit the complete implementation, no need to highlight the possible improvements separately. 

Good job! ⚡

Aminata00:
thank you @kiko-bkr !!!

PR #204, opened 2025-03-01
warmup

pahanan:


nodejs/week1/home/avg.js
```diff
@@ -0,0 +1,19 @@
+let sum = 0;
+
+if (process.argv.length === 2) {
+    console.log("Please enter a number");
+    return 0;
```
kiko-bkr, line 5: In cases where our application does not succeed, we should not return zero, as zero is the standard number indicating success. Instead use a number like 1.

And since we want this to be returned to the user's command line as the exit code, use `process.exit()` instead of `return`.

nodejs/week1/home/avg.js
```diff
@@ -0,0 +1,19 @@
...
+}
+
+for (let i = 2; i < process.argv.length; i++) {
+    for (let j = 2; j < process.argv.length; j++) {
+        if (isNaN(process.argv[j])) {
+            console.log("Please enter a number");
+            return 0;
+        }
+    }
```
kiko-bkr, line 14: This `for` loop should only be run once, as your arguments do not change over time. Therefore it is enough to run this `for` loop once before the outer `for` loop, no need to execute it for each value of `i`.

nodejs/week1/home/avg.js
```diff
@@ -0,0 +1,19 @@
...
+    sum += parseInt(process.argv[i], 10);
+}
+
+const avg = sum / (process.argv.length - 2);
```
kiko-bkr, line 18: It would make the code more readable if you created a variable that has the value `process.argv.length - 2`, and called something like `numberOfArguments`. That way you could also use it in your `for` loop conditions.

kiko-bkr (changes requested):
Hi! I left a few comments about the style of your code. The functionality is on point, but some aspects can be improved. 👍

PR #207, opened 2025-03-03
Nodejs week2/jalil

jalilhu:
Warmup Nodejs week2


nodejs/week2/search.js
```diff
@@ -0,0 +1,52 @@
+import express from "express";
+import document from "./documents.json" with { type: 'json' };
+
+export const searchRouter = express.Router();
+
+searchRouter.get("/", (req, res) => {
+    let result = document
```
ahmagdy, line 32: there is no document object in node

nodejs/week2/search.js
```diff
@@ -0,0 +1,52 @@
...
+    console.log(req.query.q)
+    let reqQuery = req.query.q
```
ahmagdy, line 34: use const instead of let

ahmagdy (approved):

PR #208, opened 2025-03-05
Nodejs week2/gayathri

GayathriVenkatraman:


nodejs/week2/app.js
```diff
@@ -0,0 +1,98 @@
...
+app.use(express.json());
+
+app.get("/", (req, res) => {
+  res.send("This is a search engine");
+});
+
+//Get all the documents, also using query
+app.get("/search", (req, res) => {
+  const { q } = req.query;
+  let documents = readDocuments();
```
Niloufar97, line 28: Reading the document file synchronously on every request can block the event loop and slow down the server. Since your data doesn't change often, it's better to either read the file once at startup and cache the data, or make the reading process asynchronous. This will improve performance and prevent blocking.

nodejs/week2/app.js
```diff
@@ -0,0 +1,98 @@
...
+  if (q) {
+    documents = documents.filter(
+      (document) =>
+        document.name.toLowerCase().includes(q.toLowerCase()) ||
+        document.description.toLowerCase().includes(q.toLowerCase())
```
Niloufar97, line 33: You should check all fields in each document (not just name and description) when implementing the search functionality.
Hint: Use` Object.values(document) `to check all values

nodejs/week2/app.js
```diff
@@ -0,0 +1,98 @@
...
+      message: "No records found",
+    });
+  }
+  res.status(200).json(documents);
+});
+
+//Get single document, by using id
+app.get("/documents/:id", (req, res) => {
+  const { id } = req.params;
+  const documents = readDocuments();
```
Niloufar97, line 48: Same here. you don't need to read your document in each endpoint when it is `fs.readFileSync`

nodejs/week2/app.js
```diff
@@ -0,0 +1,98 @@
...
+
+app.use(express.json());
+
+app.get("/", (req, res) => {
+  res.send("This is a search engine");
+});
+
+//Get all the documents, also using query
+app.get("/search", (req, res) => {
+  const { q } = req.query;
```
Niloufar97, line 27: It is a better approach to convert id to Number like this: `+req.query;`

nodejs/week2/app.js
```diff
@@ -0,0 +1,98 @@
...
+      error: "Both q and fields cannot be provided. Please use only one",
+    });
+  }
+  let documents = readDocuments();
+  if (q) {
+    documents = documents.filter(
+      (document) =>
+        document.name.toLowerCase().includes(q.toLowerCase()) ||
+        document.description.toLowerCase().includes(q.toLowerCase()) ||
+        document.price.toString().toLowerCase().includes(q.toLowerCase())
```
Niloufar97, line 75: Again: try to use `Object.values` to check all the cases

nodejs/week2/app.js
```diff
@@ -0,0 +1,98 @@
...
+      documents = documents.filter(
+        (document) =>
+          document[key] &&
+          document[key].toString().toLowerCase() ===
+            fields[key].toString().toLowerCase()
+      );
+    }
+  }
+
+  if (documents.length === 0) {
```
Niloufar97, line 90: The `fields `filter overwrites the `query `filter because both if conditions apply separately. It would be better to use `if...else `to ensure only one filtering method is applied at a time.
also it is better to check `not found `

Niloufar97 (commented):
Good job. reach out if you have any questions.

PR #211, opened 2025-03-05
Nodejs week2/farzaneh

FarzanehAhmadi:


nodejs/week2/app.js
```diff
@@ -0,0 +1,73 @@
...
+
+  if(!query){
+    return res.json(documents)
+  }
+
+  const filteredDocuments = documents.filter(doc =>
+    Object.values(doc).some(value =>
+      typeof value === 'string' && value.toLowerCase().includes(query.toLowerCase())
+    )
+  );
```
Niloufar97, line 27: What happens if a user searches for something that doesn’t exist in the data? It would be a good practice to return a `404 Not Found` response to indicate that no results were found.

nodejs/week2/app.js
```diff
@@ -0,0 +1,73 @@
...
+  }
+  let filteredDocuments = documents;
+  if(query){
+    filteredDocuments = documents.filter(doc =>
+      Object.values(doc).some(value =>
+        typeof value === "string" && value.toLowerCase().includes(query.toLowerCase())
+      )
+    );
+  }
+  if (fields) {
```
Niloufar97, line 60: The `fields `filter overwrites the `query `filter because both if conditions apply separately. It would be better to use `if...else `to ensure only one filtering method is applied at a time.
also it is better to check `not found `

Niloufar97:
Well done Farzaneh. 
I have added 2 comments. if you have any question just reach out.

PR #218, opened 2025-03-07
Nodejs week2/guzide

GuzideGuzelbey:
Thank you in advance :)


nodejs/week2/routes/search.js
```diff
@@ -0,0 +1,95 @@
...
+  }
+
+  res.status(200).json(documents);
+});
+
+//GET /documents/:id
+
+searchRouter.get("/documents/:id", (req, res) => {
+  const documents = loadDocuments();
+  const documentID = Number(req.params.id);
```
aagamezl, line 40: The number is not a good choice here, because for example the string ('1.8') will be converted to the number 1.8, which is not a valid ID.

I recommend to verify that the ID represents an integer value using [Number.isInteger()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/isInteger) after the conversion, also you must verify if the id is greater than 0 (a positive number)

nodejs/week2/routes/search.js
```diff
@@ -0,0 +1,95 @@
...
+
+  if (isNaN(documentID)) {
```
aagamezl: [Number.isNaN()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/isNaN) is a more robust version of the original global `isNaN()` function, so I recommend switching to it.

Please read the documentation I'm sharing to understand why.

nodejs/week2/routes/search.js
```diff
@@ -0,0 +1,95 @@
...
+    return JSON.parse(rawData);
+  } catch (error) {
+    console.error("Can not read JSON file:", error);
+    return [];
+  }
+};
+
+//GET /search
+searchRouter.get("/search", (req, res) => {
+  const documents = loadDocuments();
```
aagamezl: Don't upload documents in each request; move this functionality outside of the request itself, as this will optimize execution times and resources.

nodejs/week2/routes/search.js
```diff
@@ -0,0 +1,95 @@
...
+    .status(404)
+    .json({ error: `no document with this Id found: ${documentID}` });
+});
+
+// POST /search
+
+searchRouter.post("/search", (req, res) => {
+  const q = req.query.q?.trim() || null;
+  const fields = req.body.fields || null;
+  const documents = loadDocuments();
```
aagamezl: Don't upload documents in each request; move this functionality outside of the request itself, as this will optimize execution times and resources.

aagamezl (changes requested):

nodejs/week2/routes/search.js
GuzideGuzelbey: Thank you for the detailed explanation. Done!

nodejs/week2/routes/search.js
GuzideGuzelbey: Thank you for also sharing the documentation about the concepts. Done!

nodejs/week2/routes/search.js
GuzideGuzelbey: Thank you :) Done!

PR #223, opened 2025-03-11
nodejs-week3/jianxin

jianxinz233:
nodejs-w3-warmup

I tried to delete the contacts table with the given example app.js, and successfully by passing "first_name ASC; DROP TABLE contacts" parameter.


LucasIversen (approved):
All good

PR #236, opened 2025-03-19
warm up done

siderdk:
Apologies for the delay.

### DIFF
--- a/nodejs/week1/avg.js
+++ b/nodejs/week1/avg.js
@@ -1,0 +1,39 @@
+console.log("Hi I'm avg.js");
+
+const args = process.argv.slice(2);
+
+
+function checkIfNumber(arg) {
+   return !isNaN(parseFloat(arg)) && isFinite(arg);
+};
+
+function getAVG(args) {
+    const sum = args.reduce((acc, arg) => acc + parseFloat(arg), 0);
+    return sum / args.length;
+}
+
+function main(args) {
+    
+
+    if (args.length === 0) {
+        console.log("No arguments provided, please provide at least one number");
+        return;
+    } 
+    
+    if (!args.every(checkIfNumber)) {
+        console.log("Please provide only numbers");
+        return;
+    } 
+    
+    const average = getAVG(args);
+    console.log(`The average of ${args.join(", ")} is ${average}`);
+    return average;
+    } 
+
+    
+
+
+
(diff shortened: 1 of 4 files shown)


nodejs/week1/avg.js
```diff
@@ -0,0 +1,39 @@
+console.log("Hi I'm avg.js");
+
+const args = process.argv.slice(2);
+
+
+function checkIfNumber(arg) {
+   return !isNaN(parseFloat(arg)) && isFinite(arg);
+};
```
kiko-bkr, line 8: Very nice function! 🔥

nodejs/week1/avg.js
```diff
@@ -0,0 +1,39 @@
...
+
+function getAVG(args) {
+    const sum = args.reduce((acc, arg) => acc + parseFloat(arg), 0);
+    return sum / args.length;
+}
```
kiko-bkr, line 13: Nice! The casing of the name of the function is a bit odd though, it should be either `getAvg` or `getAverage` in my opinion.

nodejs/week1/avg.js
```diff
@@ -0,0 +1,39 @@
...
+    const average = getAVG(args);
+    console.log(`The average of ${args.join(", ")} is ${average}`);
+    return average;
+    } 
+
+    
+
+
+
+console.log(main(args));
```
kiko-bkr, line 37: What is the motivation for this `console.log`? You already do all the printing in the `main()` function itself, this line of code will either output an unncesessary `undefined` or a duplicate value for average, which was already printed. You can remove this.

nodejs/week1/package.json
```diff
@@ -0,0 +1,16 @@
+{
+  "name": "week1",
+  "version": "1.0.0",
+  "description": "",
+  "main": "index.js",
+  "scripts": {
+    "test": "echo \"Error: no test specified\" && exit 1",
+    "start": "node avg.js"
```
kiko-bkr, line 8: As you did not need any `npm` dependencies for your code, there is no need to make this code into an `npm` package through `npm init`. It would have been enough to just run your code via `node avg.js` in the terminal.

kiko-bkr (approved):
Nice job! I really like the way you implement. I left two small comments, but other than that LGTM!

PR #240, opened 2025-03-27
React1 week2/gayathri

GayathriVenkatraman:


GayathriVenkatraman:
Please ignore the week1 homework

react/react1/week2/components/ui/Footer.js
```diff
@@ -15,44 +68,24 @@ export const Footer = () => {
...
-            <a href="https://www.linkedin.com/">LinkedIN</a>
-          </li>
-          {/* TASK - React 1 week 2 */}
-          {/* Create a <SocialMediaItem /> component and replace all of the list items! */}
-          {/* it should accept the following props */}
-          {/* url, title, icon */}
-          {/* For the icons, you can download 1-2 social media icons for testing and put it in the /public/socialmedia/ folder */}
+          {socialMediaLinks.map((social) => (
+            <SocialMediaItem key={social.url} {...social} />
+          ))}
```
elena-bullani, line 88: Good job 💪🏻

react/react1/week2/app/destination/page.js
```diff
@@ -0,0 +1,106 @@
...
+import { useState } from "react";
+
+import styles from "@/components/destination/destination.module.css";
+import { PlanetWishlistItem } from "@/app/destination/PlanetWishlistItem";
+import { AddWishlistItem } from "@/components/destination/AddWishlistItem";
+import PlanetCard from "./PlanetCard";
+
+export const Destinations = () => {
+  const [selectedPlanets, setSelectedPlanets] = useState([]);
+  const onAddOrRemovePlanet = (name, index) => {
```
elena-bullani, line 12: Careful, there's a bug here 🐛

You should return
```javascript
return [...prevSelected, name]
```
Right now you are appending an object **containing** `name` instead of just adding the name string, therefore you are not entering the if case.

This causes problems because:
The `includes()` is looking for a string but comparing against objects
The `isSelected` prop in `PlanetCard` is using `includes()` which won't work with objects

If you run the app and keep selecting the same planet you are just appending more and more objects to that array.

elena-bullani (commented):
Hi 👋🏻 I'm Elena your mentor reviewer for this assignment ☺️

react/react1/week2/app/destination/page.js
```diff
@@ -0,0 +1,106 @@
...
+    setSelectedPlanets((prevSelected) => {
+      if (prevSelected.includes(name)) {
+        return prevSelected.filter((planet) => planet !== name);
+      } else {
+        return [...prevSelected, { name }];
```
elena-bullani: Careful, there's a bug here 🐛

You should return
```javascript
return [...prevSelected, name]
```
Right now you are appending an object **containing** `name` instead of just adding the name string, therefore you are not entering the if case.

This causes problems because:
The `includes()` is looking for a string but comparing against objects
The `isSelected` prop in `PlanetCard` is using `includes()` which won't work with objects

If you run the app and keep selecting the same planet you are just appending more and more objects to that array.

react/react1/week2/components/ui/Footer.js
GayathriVenkatraman: Thank you

react/react1/week2/app/destination/page.js
GayathriVenkatraman: Yes thanks though it's a minor defect it has lot of different effect on the outcome. Thank you for highlighting this.

GayathriVenkatraman:
Hi Elena,

Thank you so much for your time in reviewing and poviding a valuable feedback. Have a nice week ahead.

Regards,
Gayathri

PR #243, opened 2025-04-05
Nodejs week3/aminata

Aminata00:


nodejs/week3/warm-up/app.js
```diff
@@ -0,0 +1,77 @@
...
+app.use("/api", apiRouter);
+
+const contactsAPIRouter= express.Router();
+apiRouter.use("/contacts", contactsAPIRouter);
+
+contactsAPIRouter.get('/', async (req, res) => {
+ let query= knexInstance.select("*").from("contacts");
+
+ if ("sort" in req.query){
+  const order = req.query.sort.toString;
```
LucasIversen, line 30: `toString()`*

nodejs/week3/warm-up/app.js
```diff
@@ -0,0 +1,77 @@
...
+  if (order.length >0 ){
+    query = query.orderByRaw
+  }
+ }
+});
+
+  console.log("SQL", query.toSQL().sql);
+
+  try {
```
LucasIversen, line 39: This is placed outside the handler. It needs to be inside the `contactsAPIRouter.get('/',`

LucasIversen (approved):
Hi Aminata. There's a couple of things you could fix here, but you should only do it if you feel it's worth while since it's been so long since this was due :)

PR #245, opened 2025-04-06
sw1 w2 w3

StefanyRomeroV:


nodejs/week3/index.js
```diff
@@ -0,0 +1,95 @@
...
+
+// You can delete this route once you add your own routes
+apiRouter.get("/", async (req, res) => {
+  const SHOW_TABLES_QUERY =
+    process.env.DB_CLIENT === "pg"
+      ? "SELECT * FROM pg_catalog.pg_tables;"
+      : "SHOW TABLES;";
+  const tables = await knex.raw(SHOW_TABLES_QUERY);
+  res.json({ tables });
+});
```
kiko-bkr: There is no need for this code anymore, as the comment states.

nodejs/week3/index.js
```diff
@@ -0,0 +1,95 @@
...
+app.use("/api", apiRouter);
+
+
+app.listen(process.env.PORT, () => {
+  console.log(`API listening on port ${process.env.PORT}`);
+});
+
+app.get("/my-route", (req, res) => {
+  res.send("Hi friend");
+});
```
kiko-bkr: This is also not needed.

nodejs/week3/index.js
```diff
@@ -0,0 +1,95 @@
...
+
+//*/future-meals	Respond with all meals in the future (relative to the when datetime)//*
+app.get("/future-meals", async (req, res) => {
+
+  const GET_MEALS_QUERY = "SELECT * FROM meals WHERE when > NOW();";
+  const meals = await knex.raw(GET_MEALS_QUERY);
+  res.json({ meals });
+}
+);
```
kiko-bkr: I like the comments that explain what each endpoint does. One small detail that you are not accounting for is that you can encounter a database error. (for example, if the connection closes because of issues outside of your app)

For these cases, it would be best to use a `try - catch` block, where in the `catch` block you let the user know with a status code of `500` that an internal server error occurred. 🤓

nodejs/week3/index.js
```diff
@@ -0,0 +1,95 @@
...
+app.get("/my-route", (req, res) => {
+  res.send("Hi friend");
+});
+
+//*/future-meals	Respond with all meals in the future (relative to the when datetime)//*
+app.get("/future-meals", async (req, res) => {
+
+  const GET_MEALS_QUERY = "SELECT * FROM meals WHERE when > NOW();";
+  const meals = await knex.raw(GET_MEALS_QUERY);
+  res.json({ meals });
```
kiko-bkr: The variable meals in this case is an array of two objects: the first element is the actual result, (array of databae rows as JS objects) and the second one is a string that represents the schema of the table. You should only return the first element: the array. 
```suggestion
  const [meals, schema] = await knex.raw(GET_MEALS_QUERY);
  res.json({ meals });
```

nodejs/week3/index.js
```diff
@@ -0,0 +1,95 @@
...
+);
+//* all meals	Respond with all meals sorted by ID//*
+
+app.get("/all-meals", async (req, res) => {
+  
+    const GET_MEALS_QUERY = "SELECT * FROM meals ORDER BY id;";
+    const meals = await knex.raw(GET_MEALS_QUERY);
+    res.json({ meals });
+  }
+  );
```
kiko-bkr: Same comments apply for these endpoints.

nodejs/week3/index.js
```diff
@@ -0,0 +1,95 @@
...
+  //*respond with the last meal meaning with the maximum id*//
+
+  app.get("/last-meal", async (req, res) => {
+    
+      const GET_MEALS_QUERY = "SELECT * FROM meals ORDER BY id DESC LIMIT 1;";
+      const meals = await knex.raw(GET_MEALS_QUERY);
+      res.json({ meals });
+    }
+    );
```
kiko-bkr: And the same applies to these endpoints. Moreover, for these endpoints, the user expects to receive a single element, not an array. This means that if the length of `meals[0]` is 0, then we should return a `404` not found error to the user.

kiko-bkr (changes requested):
Well done Julie! Your code is organized, but is lacking some error handlng. Please extend your endpoints with the error handling cases, and make sure to supply the correct return types. ⚡

PR #246, opened 2025-04-07
Nodejs/week3/stefany

StefanyRomeroV:


nodejs/week3/src/routers/meals.js
```diff
@@ -0,0 +1,232 @@
...
+    console.error("Error fetching meals:", error);
+    res.status(500).json({ error: "Failed to fetch meals" });
+  }
+};
+//GET api/meals with query parameters
+mealsRouter.get("/", async (req, res) => {
+  try {
+    let meals;
+    if (!req.query.availableReservations) {
+      meals = await getAllMeals();
```
LucasIversen, line 20: I see what you're doing. It's sometimes easier to filter on an array in javascript than expanding the knex query. However, this also means that we always fetch every single meal, even when we maybe only need to return a couple of them. Now imagine we have 10000 meals, that's a lot of data to fetch every time. Therefore, it's better practice to make the query only fetch what you need.

nodejs/week3/src/routers/meals.js
```diff
@@ -0,0 +1,232 @@
...
+            return 0;
+          })
+      }
+      else {
+        return res.status(400).json({ error: "Invalid sortKey value" });
+      }
+    };
+
+    //sortDir
+    if (req.query.sortDir) {
```
LucasIversen, line 108: The sortDir should only be used if the sortKey is also used.

nodejs/week3/src/routers/meals.js
```diff
@@ -0,0 +1,232 @@
...
+    res.status(500).json({ error: "Failed to fetch meal" });
+  }
+}
+);
+//PUT/api/meals/:id
+mealsRouter.put("/:id", async (req, res) => {
+  try {
+    const mealId = +req.params.id;
+    const reqBody = req.body;
+    const meal = await knex("meals").where({ id: mealId }).first();
```
LucasIversen, line 152: We can avoid this fetch by checking the updatedMealNum instead

nodejs/week3/src/routers/reviews.js
```diff
@@ -0,0 +1,125 @@
...
+        console.error('Error adding review:', error);
+        res.status(500).json({ error: 'Failed to add review' });
+
+    }
+}
+);
+//get/reviews/:id
+const getReviewById = async (id) => {
+    try {
+        const allReviews = await getAllReviews();
```
LucasIversen, line 71: Here we also fetch all the reviews only to return a single one. Let's instead only fetch the meal with the id instead

LucasIversen (changes requested):
Hi Stefany. I've got a few comments, mainly regarding how we fetch data. If anything needs further explaining, you can always write me on slack 👌

